PR #162, opened 2018-02-21
Allow UTM params

davecozens:


james-siteclick:
Would it be OK to set a reasonable-sounding max length constraint on these?

PR #187, opened 2018-03-01
Feat/combine js

james-siteclick:


davecozens:
As discussed, waiting until after public beta release

PR #362, opened 2018-07-30
Feature: Handle user role with returns permission

swaff:
Updates the permissions object so that an external user can have a
returns permission.

This may be a primary_user, or a user who has been allocated returns
permission by the primary user.

### DIFF
--- a/src/lib/permissions.js
+++ b/src/lib/permissions.js
@@ -1,6 +1,9 @@
 'use strict';
 
-const { intersection } = require('lodash');
+// Using lodash functions over some native functions because
+// they are forgiving to undefined.
+const { size, find, includes, intersection, get } = require('lodash');
+
 /**
  * Permissions module
  *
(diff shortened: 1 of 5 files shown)


src/lib/permissions.js
```diff
@@ -21,33 +24,68 @@ const { intersection } = require('lodash');
...
+  return isExternal(scope) && size(roles) === 1 && isPrimaryUser(roles);
 };
 
 const canViewMutlipleLicences = (scope = [], roles = []) => {
   return isExternal(scope) && countRoles(roles, ['user', 'primary_user']) > 1;
 };
 
+const getUserRole = roles => find(roles, isUserRole);
+
+const canPerformReturns = (scope = [], roles = []) => {
```
james-siteclick: const canPerformReturns = (scope = [], roles = []) => {
  const returnsRole = roles.find((role) => {
    const isPrimaryUser = role.role === 'primary_user';
    const isUserWithPermission = role.role === 'user' && (get(role, 'permissions.returns') === true);
    return isPrimaryUser || isUserWithPermission;
  });
  return isExternal(scope) && !!returnsRole;
};

PR #427, opened 2018-10-10
Adding JSON schema resolver and form generator for WR22 conditions

james-siteclick:


src/modules/abstraction-reform/lib/form-generator.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+/**
+ * Guess label given a field name
+ * Converts underscore to space and converts to sentence case
+ * @param {String} str - field name, snake case
+ * @return {String} label
+ */
+const guessLabel = (str) => {
+  return sentenceCase(str.replace(/_/, ' '));
```
swaff: Should this be `return sentenceCase(str.replace(/_/g, ' '));` to replace all the underscores?

PR #542, opened 2018-12-12
Feature/water 1526 json schema validation adapter

swaff:


src/lib/forms/validationAdapters/json-schema.js
```diff
@@ -0,0 +1,87 @@
...
+  };
+  return formattedError;
+};
+
+/**
+ * Copies the errors to the local form field level if
+ * there is an error from that field
+ */
+const applyLocalErrors = (form, errors) => {
+  form.fields.forEach(field => {
```
james-siteclick: Fields can contain child fields (in the case of conditional fields) so may need to use mapFields function to apply errors

PR #543, opened 2018-12-12
Create add action to add WR22 data element to arData array within lic…

james-siteclick:
…ence data object


src/modules/abstraction-reform/lib/action-creators.js
```diff
@@ -161,6 +166,62 @@ const createEditAddress = (data, user, id) => {
   };
 };
 
+/**
+ * Creates a new data object - e.g. for WR22 data
+ * @param {String} id - GUID of new data point
```
swaff: Comments are out of sync with the function signature

PR #545, opened 2018-12-13
Feat/gov uk design system poc 2

james-siteclick:


src/assets/sass/v2/_typography.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.sub-head {
+  font-size: 30px;
+  color: #6f777b;
+  margin: 0;
+  padding: 0;
+}
+
+// ========== ========== ========== ========== ==========
```
swaff: From here to the end of the file is also present in the _styles.scss file. Is this correct?

Also I think we should probably avoid styling by id.

gulpfile.js
```diff
@@ -103,7 +103,8 @@ gulp.task('sass', () => {
         'govuk_modules/govuk_template_mustache/assets/stylesheets',
         'govuk_modules/govuk-elements-sass',
         'govuk_modules/govuk-elements-sass/public/sass',
-        'govuk_modules'
+        'govuk_modules',
+        'node_modules'
```
swaff: Could this be more specific?

src/lib/view-engine/index.js
```diff
@@ -0,0 +1,17 @@
...
+  engines: {
+    html: handlebarsEngine,
+    njk: nunjucksEngine
+  },
+  path: './src/views',
+  layoutPath: './src/views/govuk_template_mustache/layouts',
+  partialsPath: './src/views/partials',
+  layout: 'govuk_template',
+  context: defaultContext,
+  isCached: process.env.NODE_ENV === 'production',
```
swaff: Should we add pre-prod here as well to have the environments as similar as possible?

PR #568, opened 2019-01-08
Feat: Use accessible autocomplete for long selects

swaff:
When a select element contains more than 200 items and the user permits
client side JS, the select is replaced with an accessible autocomplete
component.

Adds the govuk-body class to the form-group that contains the select to
allow the style to cascade into the list of options that replace the
select options.

### DIFF
--- a/src/views/nunjucks/abstraction-reform/edit-data.njk
+++ b/src/views/nunjucks/abstraction-reform/edit-data.njk
@@ -4,9 +4,6 @@
 {% from "forms/footer.njk" import formFooter %}
 {% from "forms/widget.njk" import formWidget %}
 
-
-
-
 {% block content %}
 
   {{ formErrorSummary(form) }}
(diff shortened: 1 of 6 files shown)


src/views/nunjucks/abstraction-reform/edit-data.njk
```diff
@@ -39,5 +33,22 @@
   integrity="sha256-FgpCb/KJQlLNfOu91ta32o/NMZxltwRo8QtmkMRdAu8="
   crossorigin="anonymous"></script>
 <script src="/public/javascripts/json-forms-toggle.js"></script>
-
+<script src="/public/javascripts/accessible-autocomplete.min.js"></script>
+<link href="/public/stylesheets/accessible-autocomplete.min.css" rel="stylesheet" />
+<script nonce={{nonces.script}}>
```
james-siteclick: If this inline code chunk is the same as the one in above, would it be worth moving it to its own JS file?

PR #586, opened 2019-01-18
Creates in the UI the ability to CSV preview and send final return re…

james-siteclick:
…minders.  Refactors CSV download for abstraction reform report.


src/modules/returns-notifications/lib/helpers.js
```diff
@@ -1,9 +1,36 @@
...
 const getUniqueLicenceNumbers = (returns) => {
   return uniq(returns.map(row => row.licence_ref));
 };
 
+/**
+ * Gets the end date of the current cycle
+ * @param  {String} date - Optional reference date
+ * @return {String}      - end date of the current cycle
+ */
+const getCurrentCycleEndDate = (date) => {
```
swaff: This could go into the water-abstracton-helpers module maybe?

PR #596, opened 2019-01-29
Feat: Adds the user status page for internal users

swaff:
WATER-1833

First part of story 1833 which is a new user status page that shows the
licences, verification and companies that a user is associated with.


src/assets/sass/v2/blocks/_space.scss
```diff
@@ -0,0 +1,6 @@
+.space--none {
```
james-siteclick: Would govuk-!-margin-bottom-0 from the kit work instead of adding our own space classes?

PR #600, opened 2019-01-31
Feat: Adds registered user to licence detail page

swaff:
WATER-1833

If the user is an admin, changes the licence detail page to show that a
licence is either unregistered, or who the licence is registered to.

 - Adds the user name to the licence page if the licence has a primary
user.
 - Updates the security code display to show the recipient address if a
verification code has been sent.
 - Remove some unused CSS classes.

### DIFF
--- a/src/lib/connectors/water-service/licences.js
+++ b/src/lib/connectors/water-service/licences.js
@@ -1,26 +1,39 @@
-const rp = require('request-promise-native').defaults({ proxy: null, strictSSL: false });
+const serviceRequest = require('../service-request');
+const { partialRight } = require('lodash');
 
 const get = (documentId, tail) => {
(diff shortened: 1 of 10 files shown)


src/lib/connectors/water-service/licences.js
```diff
@@ -1,26 +1,39 @@
...
+const getLicenceUsersByDocumentId = partialRight(get, 'users');
+
+const getLicencePrimaryUserByDocumentId = async documentId => {
+  try {
+    const userResponse = await getLicenceUsersByDocumentId(documentId);
+
+    if (!userResponse.error) {
+      const users = userResponse.data || [];
+      return users.find(user => user.roles.includes('primary_user'));
+    }
```
james-siteclick: Is it OK that userResponse.error is ignored here?

PR #604, opened 2019-02-01
Fix: User status page visual bugs

swaff:
WATER-1833

 - Use proper plurals for licence and licences in the user status page
 - Allow email addresses to overflow onto the next line because they can
be too long for a single line in mobile view.


src/views/nunjucks/macros/pluralise.njk
```diff
@@ -0,0 +1,3 @@
+{% macro pluralise(collection, singular, plural) %}
```
james-siteclick: In my latest PR I've also created a plural implementation using a Nunjucks filter - perhaps we can have a quick look at this together

PR #607, opened 2019-02-05
Feat: Adds a communication viewer page

swaff:
WATER-1891

Adds a page that shows the message content that has been previously
sent.

 - Updates marked module to latest version to use option to remove header
ids that are added to the output html.

 - Replace pluralise macro with pluralize filter

 - Adds the back link to the communication detail page

 - Fixes an issue with the dates in the communications list.

### DIFF
--- a/src/modules/view-licences/controller.js
+++ b/src/modules/view-licences/controller.js
@@ -230,7 +230,7 @@ const getLicenceCommunication = async (request, h) => {
   const licence = response.data.licenceDocuments.find(doc => doc.documentId === documentId);
 
   const validationError = validateLicenceCommunicationResponses(request, licence);
-  if (validationError) return validationError;
+  if (validationError) throw validationError;
 
   const viewContext = {
     ...request.view,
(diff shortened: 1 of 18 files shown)


src/modules/view-licences/routes-admin.js
```diff
@@ -99,6 +99,11 @@ const getLicenceGaugingStationAdmin = {
   path: '/admin/licences/{licence_id}/station/{gauging_station}'
 };
 
+const getLicenceCommunication = {
```
james-siteclick: Does this route definition include internal scope?

src/modules/view-licences/controller.js
```diff
@@ -191,10 +192,65 @@ const getLicence = async (request, h) => {
...
+};
+
+const getLicenceCommunication = async (request, h) => {
+  const { communicationId, documentId } = request.params;
+  const response = await communicationsConnector.getCommunication(communicationId);
+
+  const licence = response.data.licenceDocuments.find(doc => doc.documentId === documentId);
+
+  const validationError = validateLicenceCommunicationResponses(request, licence);
+  if (validationError) return validationError;
```
james-siteclick: I'd prefer to stick with throw here for now, as that's the pattern used elsewhere in the service

PR #610, opened 2019-02-06
Fix: Adds screen reader aware text in comms table

swaff:
WATER-1892

Updates the 'type' column text in the communications table to include
additional context for screen reader user agents.


src/assets/sass/v2/blocks/_text.scss
```diff
@@ -1,3 +1,16 @@
 .break-word {
   overflow-wrap: break-word;
 }
+
+// Screen-reader only text
+// @see https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/
+.sr-only {
```
james-siteclick: I think there is already `govuk-visually-hidden` class for this

PR #627, opened 2019-02-25
Feat/water 1432

james-siteclick:


src/modules/view-licences/helpers.js
```diff
@@ -1,10 +1,11 @@
-const { LicenceNotFoundError } = require('./errors');
-const CRM = require('../../lib/connectors/crm');
+// const { LicenceNotFoundError } = require('./errors');
+// const CRM = require('../../lib/connectors/crm');
```
swaff: 💣

src/modules/view-licences/helpers.js
```diff
@@ -1,10 +1,11 @@
...
 const Permit = require('../../lib/connectors/permit');
 const LicenceTransformer = require('../../lib/licence-transformer/');
 const waterConnector = require('../../lib/connectors/water');
 const { find, has } = require('lodash');
 const Boom = require('boom');
+// const permissions = require('../../lib/permissions');
```
swaff: 🔥

test/lib/hapi-plugins/auth-credentials.js
```diff
@@ -0,0 +1,158 @@
...
+    };
+    sandbox.stub(idmConnector, 'getUserByEmail');
+    sandbox.stub(logger, 'error');
+  });
+
+  afterEach(async () => {
+    sandbox.restore();
+  });
+
+  test('it should return h.continue on unauthenticated routes', async () => {
```
swaff: With the test functions, it seems better to me not to prefix with 'it should' because when the test prints it will read "auth credentials plugin handler it should return h.continue on unauthenticated routes"

I think `test('returns h.continue on unauthenticated routes')` works better leading to "auth credentials plugin  returns h.continue on unauthenticated routes" being printed.

PR #638, opened 2019-03-05
Fix: Prevent redirect of users to welcome page

swaff:
WATER-1955

Fixes a bug where the user could end up on a page that is not the
company selector without having selected a company. This mean licence
lookups where not authorised so the user was redirected to the welcome
page.

This change updates the `auth-credentials` plugin to add more content to
the request under the `defra` property (to avoid collisions). This data
is then used in a `preResponse` handler to redirect the user either to
the `select-company` route if the user has companies, or the
`add-licences` route if they don't.

Also hides the company switcher if there is no company name selected
preventing it showing if the user has not selected a company but wishes
to update a password.

### DIFF
--- a/src/lib/hapi-plugins/company-selection.js
+++ b/src/lib/hapi-plugins/company-selection.js
@@ -1,0 +1,51 @@
+/**
+ * This plugin will redirect a user to the select-company route
+ * of the user has not selected a company, but does have entity roles
+ * and therefore a company.
+ *
+ * If the user has no company then they will be redirected to the
+ * add-licences route.
+ *
+ * Both of the above redirections will only take place if the user is
+ * attempting to access a route that is configured with an access scope.
+ */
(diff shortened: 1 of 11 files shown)


src/lib/hapi-plugins/company-selection.js
```diff
@@ -0,0 +1,51 @@
...
+ * add-licences route.
+ *
+ * Both of the above redirections will only take place if the user is
+ * attempting to access a route that is configured with an access scope.
+ */
+const { get } = require('lodash');
+const SELECT_COMPANY_PATH = '/select-company';
+const ADD_LICENCES_PATH = '/add-licences';
+
+const shouldRedirect = request => {
```
james-siteclick: const isExternalUser = permissions.isExternal(request);
const companyId = permissions.getCompany(request);

PR #649, opened 2019-03-07
Add xmlUsers check, add xml upload link to return page for xmlUsers

DGajewska:
WATER-1963

Add xmlUser flag to returnsViewData which checks if the licences are done by xml upload and if the user is an externalReturns user. Xml upload link displayed if criteria is met.


src/modules/returns/lib/helpers.js
```diff
@@ -96,6 +94,30 @@ const getLicenceReturns = async (licenceNumbers, page = 1, isInternal = false) =
...
 };
 
+/**
+ * Checks whether user uses XML Upload for a list of licence numbers
+ * @param {Array} licenceNumbers to check if isUpload flag is true
+ * @param {Boolean} isInternal to specify user type
+ * @return {Boolean} if user has XML Upload functionality
+ */
+const isXmlUpload = async (licenceNumbers) => {
+  const filter = {
```
james-siteclick: I think we might need to add some additional filter params here:
```
const filter = {
status : 'due',
end_date : { $gte : '2018-10-31' }
'metadata->>isCurrent' : 'true'
...
```

src/modules/returns/lib/helpers.js
```diff
@@ -96,6 +94,30 @@ const getLicenceReturns = async (licenceNumbers, page = 1, isInternal = false) =
...
+    'metadata->>isUpload': 'true',
+    'licence_ref': { '$in': licenceNumbers }
+  };
+
+  const requestPagination = { 'page': 1, 'perPage': 1 };
+  const columns = ['return_id'];
+
+  const { error, pagination } = await returns.findMany(filter, {}, requestPagination, columns);
+  if (error) {
```
james-siteclick: We could shorten this a little if we imported `throwIfError` from hapi-pg-rest-api:
```
throwIfError(error);
return pagination.totalRows > 0
```

src/modules/returns/lib/helpers.js
```diff
@@ -96,6 +94,30 @@ const getLicenceReturns = async (licenceNumbers, page = 1, isInternal = false) =
   return { data, pagination };
 };
 
+/**
+ * Checks whether user uses XML Upload for a list of licence numbers
+ * @param {Array} licenceNumbers to check if isUpload flag is true
+ * @param {Boolean} isInternal to specify user type
+ * @return {Boolean} if user has XML Upload functionality
```
james-siteclick: As it's an async function it returns a promise.  Having had a quick look at JSDoc docs, I think this should probably be:
```
@return {Promise<boolean>}
```
Which I haven't been doing...  I have only been doing @return {Promise}

src/modules/returns/lib/helpers.js
```diff
@@ -96,6 +94,30 @@ const getLicenceReturns = async (licenceNumbers, page = 1, isInternal = false) =
   return { data, pagination };
 };
 
+/**
+ * Checks whether user uses XML Upload for a list of licence numbers
+ * @param {Array} licenceNumbers to check if isUpload flag is true
+ * @param {Boolean} isInternal to specify user type
```
james-siteclick: I think this param has been removed

src/modules/returns/lib/helpers.js
```diff
@@ -237,11 +259,14 @@ const getReturnsViewData = async (request) => {
 
   const documents = await getLicenceNumbers(request, filter);
   const licenceNumbers = documents.map(row => row.system_external_id);
+  const xmlUpload = await isXmlUpload(licenceNumbers);
+  const externalReturns = await isExternalReturns(request);
```
james-siteclick: Although this works with an `await`, `isExternalReturns` is synchronous so the `await` is not needed

PR #679, opened 2019-03-25
Feat/water 1975 merge

james-siteclick:


src/assets/sass/blocks/_returns.scss
```diff
@@ -0,0 +1,14 @@
+.licence-number-heading {
```
james-siteclick: Is this CSS class used anywhere?

src/assets/sass/blocks/_returns.scss
```diff
@@ -0,0 +1,14 @@
...
+  color: #6f777b;
+
+}
+
+.text-medium {
+  font-size: 1.25em;
+  font-weight: lighter;
+}
+
+.skip-button {
```
james-siteclick: Also is this CSS class used anywhere?

src/modules/returns/controllers/edit.js
```diff
@@ -353,18 +355,24 @@ const getConfirm = async (request, h) => {
   const lines = getLinesWithReadings(data);
   const form = confirmForm(request, data, `/return/confirm`);
 
+  const isReadings = get(data, 'reading.method') === 'oneMeter';
+  const endReadingKey = findLastKey(get(data, 'meters[0].readings'), key => key > 0);
```
james-siteclick: Would this work here:
```
const endReading = last(get(data, 'meters[0].readings', []))
```

src/modules/returns/lib/return-helpers.js
```diff
@@ -310,7 +294,25 @@ const applyMeterReadings = (data, formValues) => {
...
+
+const applyMeterReset = (data, formValues) => {
+  const { meterReset } = formValues;
+  const updated = cloneDeep(data);
+
+  if (meterReset) {
+    set(updated, 'reading.method', 'abstractionVolumes');
+    const meters = updated.meters || [];
+    for (let meter of meters) {
+      delete meter.readings;
```
james-siteclick: Just checking - do we need to delete these properties or just set them back to null, empty arrays etc.

test/modules/returns/controllers/edit.js
```diff
@@ -694,6 +696,50 @@ experiment('edit controller', () => {
     });
   });
 
+  experiment('getMeterReset', () => {
+    test('it should take you to water/returns/internal/form page with returns data', async () => {
```
james-siteclick: I misread this thinking the route was doing a redirect, could we maybe say
`it should render...`

src/assets/sass/blocks/_returns.scss
```diff
@@ -0,0 +1,14 @@
+.licence-number-heading {
+  color: #6f777b;
+
+}
+
+.text-medium {
```
james-siteclick: Just wondering what this class does

PR #685, opened 2019-03-28
Fix: Prevent empty inset text container

swaff:
WATER-2014

If there are no meter details, or there is no estimate for the
abstraction, then don't show the inset text container.

Also fixes an issue with the meter details being override to a 1x
multiplier.

### DIFF
--- a/src/views/nunjucks/returns/macros/confirm-page-inset-text.njk
+++ b/src/views/nunjucks/returns/macros/confirm-page-inset-text.njk
@@ -1,61 +1,49 @@
 {% from "inset-text/macro.njk" import govukInsetText %}
 
 {%macro confirmPageInsetText(return, endReading)%}
-{% set html %}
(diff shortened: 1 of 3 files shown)


src/views/nunjucks/returns/macros/confirm-page-inset-text.njk
```diff
@@ -1,61 +1,49 @@
...
+        </div>
+      {% endif %}
+    {% endif %}
+
+    {%if return.reading.method == 'oneMeter' %}
+      <div class="meta__row">
+        <dd class="meta__value">Start Reading {{ return.meters[0].startReading }}</dd>
+      </div>
+      <div class="meta__row">
+        <dd class="meta__value">End Reading {{ endReading }}</dd>
```
james-siteclick: Should this be `End reading` - usual GOV.UK style is sentence case

PR #720, opened 2019-04-04
Add deletion of meter readings if submitting abstraction volumes to a…

DGajewska:
…pplyCleanUp

WATER-2063

Was receiving 500 error on submission of abstraction volumes if meter readings existed in data. Add deletion of meter readings if reading method is abstraction volumes.

### DIFF
--- a/src/modules/returns/lib/return-helpers.js
+++ b/src/modules/returns/lib/return-helpers.js
@@ -468,8 +468,11 @@ const applyCleanup = (data, request) => {
     }
     // Delete startReading and meter readings if submitting abstractionVolumes
     if (get(updated, 'reading.method') === 'abstractionVolumes') {
-      delete updated.meters[0].startReading;
-      delete updated.meters[0].readings;
+      updated.meters = updated.meters.map(meter => {
+        delete meter.startReading;
+        delete meter.readings;
+        return meter;
+      });
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/modules/returns/lib/return-helpers.js
```diff
@@ -466,6 +466,11 @@ const applyCleanup = (data, request) => {
         };
       });
     }
+    // Delete startReading and meter readings if submitting abstractionVolumes
+    if (get(updated, 'reading.method') === 'abstractionVolumes') {
+      delete updated.meters[0].startReading;
```
james-siteclick: meters[0] may be undefined.
could consider:
```
updated.meters = updated.meters.map(meter => {
  delete meter.startReading;
  delete meter.readings;
})
```

test/modules/returns/lib/return-helpers.js
```diff
@@ -891,6 +898,18 @@ experiment('applyCleanup', () => {
...
+      expect(result.meters[0].readings).to.not.exist();
+      expect(result.meters[0].startReading).to.not.exist();
+    });
+
+    test('does not remove meter readings and startReading if method is not abstractionVolumes', async () => {
+      const result = applyCleanup(createReturn(), createRequest());
+      expect(result.meters[0].readings).to.exist();
+      expect(result.meters[0].startReading).to.exist();
+    });
   });
```
james-siteclick: Maybe add a test here for when meters is an empty array

james-siteclick:
:shipit:

PR #724, opened 2019-04-10
Add FAO information to letter when adding new licence

DGajewska:
WATER-1317

Add FAO page after selectAddress page. Add FAO to address lines in Notify connector. Add new nunjucks pages. Create for-attention-of form.


src/modules/add-licences/controller.js
```diff
@@ -350,26 +407,26 @@ async function postAddressSelect (request, h) {
...
-    const viewContext = await getAddressSelectViewContext(request, verification, addressLicence);
+    const viewContext = await getAddressSelectViewContext(request, verification, addressLicence, fao);
 
-    return h.view('water/licences-add/verification-sent', viewContext);
-  } catch (err) {
-    if (err.name === 'InvalidAddressError') {
-      return h.redirect('/select-address?error=invalidAddress');
-    }
-    throw err;
+    return h.view('nunjucks/licences-add/verification-sent.njk', viewContext, { layout: false });
```
james-siteclick: For the new and improved Nunjucks views, if we could make the folder name match the module name

src/modules/add-licences/controller.js
```diff
@@ -321,6 +318,17 @@ const getLicence = async documentId => {
   return data;
 };
 
+const updateFormForInvalidAddress = (form) => {
```
james-siteclick: We can remove the need for bespoke validation by creating a dynamic schema for both forms

src/modules/add-licences/forms/for-attention-of.js
```diff
@@ -0,0 +1,47 @@
...
+  f.fields.push(fields.hidden('csrf_token', {}, csrfToken));
+
+  return f;
+};
+
+/**
+ * Gets validation schema for log receipt form
+ * @return {Object} Joi validation schema
+ */
+const getSchema = () => {
```
james-siteclick: I think we can get Joi to check the IDs are part of a valid set for as as so:
```
/**
 * Gets validation schema for log receipt form
 * @return {Object} Joi validation schema
 */
const getSchema = (selectedIds) => {
  return {
    csrf_token: Joi.string().guid().required(),
    selectedAddressId: Joi.string().guid().required().valid(selectedIds),
    fao: Joi.string().max(32).allow('')
  };
};
```

src/modules/add-licences/forms/select-address.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    choices: lines
+  }, selectedAddressId));
+
+  f.fields.push(fields.button(null, { label: 'Continue' }));
+  f.fields.push(fields.hidden('csrf_token', {}, csrfToken));
+
+  return f;
+};
+
```
james-siteclick: We can introduce a dynamic schema for this form too along the lines of:
```
const selectAddressSchema = (licences) => {
  const documentIds = licences.map(licence => licence.document_id);
  return {
    csrf_token: Joi.string().guid().required(),
    selectedAddressId: Joi.string().guid().required().valid(documentIds)
  };
};
```

src/modules/add-licences/forms/select-address.js
```diff
@@ -0,0 +1,53 @@
...
+  const f = formFactory(action);
+
+  const lines = getLines(licences);
+
+  f.fields.push(fields.radio('selectedAddressId', {
+    errors: {
+      'any.required': {
+        message: 'Select an address'
+      },
+      'invalidAddress': {
```
james-siteclick: Once we've used Joi's valid() validator, this would become `any.allowOnly`

src/views/nunjucks/form.njk
```diff
@@ -0,0 +1,11 @@
+{% extends "./nunjucks/layout.njk" %}
+{% from "nunjucks/macros/forms/index.njk" import formRender, formErrorSummary %}
+{% from "nunjucks/macros/title.njk" import title %}
+
+{% block content %}
+  <div class="govuk-grid-row">
```
james-siteclick: I think this needs an extra div with a column class to render with the correct margin at mobile sizes

PR #727, opened 2019-04-12
WATER-2087

swaff:
WATER-2087

Creates new page that starts the processing of the returns reminders.

 - Adds new link to the reports and notifications page
 - Create new page to start the process
 - Adds placeholder route/controller for the waiting page
 - Renames existing `send-reminders` form to `send-final-reminders`

Implementation of waiting page to be done in story 2089

### DIFF
--- a/src/lib/connectors/water-service/batch-notifications.js
+++ b/src/lib/connectors/water-service/batch-notifications.js
@@ -1,0 +1,24 @@
+const serviceRequest = require('../service-request');
+const config = require('../../../../config');
+
+const getBaseUrl = () => `${config.services.water}/batch-notifications`;
+
+const prepareReturnsReminders = (issuer, excludeLicences) => {
+  const url = `${getBaseUrl()}/prepare/returnReminder`;
+  return serviceRequest.post(url, {
+    body: {
+      issuer,
+      data: {
+        excludeLicences
+      }
+    }
+  });
+};
+
+const sendReturnsReminders = eventId => {
+  const url = `${getBaseUrl()}/send/${eventId}`;
+  return serviceRequest.post(url);
+};
+
+exports.prepareReturnsReminders = prepareReturnsReminders;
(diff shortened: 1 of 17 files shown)


src/lib/connectors/water-service/batch-notifications.js
```diff
@@ -0,0 +1,24 @@
...
+    body: {
+      issuer,
+      data: {
+        excludeLicences
+      }
+    }
+  });
+};
+
+const sendReturnsReminders = eventId => {
```
james-siteclick: I think I need to make a change to the water service so that the issuer is also sent with this request, and this is checked against the issuer of the prepare event above.

src/views/nunjucks/returns-notifications/reminders.njk
```diff
@@ -0,0 +1,8 @@
+{% from "button/macro.njk" import govukButton %}
```
james-siteclick: Dana created a page at /src/views/nunjucks/form.njk for generic form pages that just contain a form object and page title which could be used here.  not sure if you have it in your branch though

src/modules/returns-notifications/forms/send-reminders.js
```diff
@@ -1,23 +1,27 @@
...
-  const action = `/admin/returns-notifications/final-reminder`;
+  const action = `/admin/returns-notifications/reminders`;
 
   const f = formFactory(action);
-  f.fields.push(fields.button(null, { label: 'Send reminders' }));
   f.fields.push(fields.hidden('csrf_token', {}, csrfToken));
+  f.fields.push(fields.text('excludeLicences', {
+    label: 'Exclude licences',
+    multiline: true,
+    hint: 'Separate licence numbers with a comma or new line'
```
james-siteclick: If this field used the mapper:
```
mapper : 'licenceNumbersMapper'
```
The values would be mapped to an array without further processing

PR #730, opened 2019-04-15
Fix: Use smaller font size on 'Return Submitted' page

DGajewska:
WATER-2048

Extend govuk body and body-l styles to white lettering on coloured panel


src/assets/sass/v2/blocks/_panel.scss
```diff
@@ -0,0 +1,12 @@
+.panel-body-title--white-text {
```
james-siteclick: Since we are diverging from the design of the standard component, we could create a new component of our own that extends the original but adds a modifier to allow smaller text in the body section:
```
.panel {
  @extends .gov-uk-panel;
}
.panel--confirmation {
  @extends .gov-uk-panel--confirmation;
}
.panel__title {
  @extends .gov-uk-panel__title;
}
.panel__body {
  @extends .gov-uk-panel__body;
}
.panel__body--small {
  font-size: 1rem; 
}
```

src/views/nunjucks/returns/submitted.njk
```diff
@@ -3,16 +3,24 @@
 
 {% block content %}
 
+{% set titleHtml %}
```
james-siteclick: We need to make sure we are using a heading element for the title, probably H1

PR #769, opened 2019-05-21
Pass isFinal flag, to deal with split logs, when creating CSV return …

DGajewska:
…files

WATER-2012

Pass isFinal flag to helpers function when checking which rows should be edited in the creation of the CSV return file.


DGajewska:
Superceded by PR # 995

PR #771, opened 2019-05-22
Add 'Upload more returns' link to submission page

DGajewska:
WATER-2129

Get xmlUser data on submission page and show link if xmlUser is true. Link takes user back to upload page.


src/external/modules/returns/controllers/upload.js
```diff
@@ -275,12 +276,18 @@ const postSubmit = async (request, h) => {
  * Page to render a success message
  */
 const getSubmitted = async (request, h) => {
+  console.log('inside getSubmitted function');
```
james-siteclick: Would it be possible to remove all the `console.logs`?

PR #774, opened 2019-05-23
Update licence points, purposes and contact details to Nunjucks

DGajewska:
WATER-2136

Update licence points, purposes and contact details to Nunjucks. Create ngrPoints Nunjucks filter. Delete obsolete html view files. Add custom styling for summary list component.


src/external/views/nunjucks/view-licences/purposes.njk
```diff
@@ -0,0 +1,113 @@
...
+            {% endfor %}
+          </dl>
+        {% else %}
+        <dd class="summary-list__value--nested-list">
+          <details>
+            <summary>View your {{ purpose.points.length }} abstraction points for {{ purpose.name | lower }}</summary>
+            <dl class="govuk-summary-list summary-list--nested-list">
+              {% for abstractionPoint in purpose.points %}
+                  <div class="govuk-summary-list__row">
+                    <dd class="{% if loop.index == purpose.points.length %} summary-list__value--no-bottom-border {%else%} govuk-summary-list__value  {%endif%}">
```
james-siteclick: Would loop.last work here?

src/external/views/nunjucks/view-licences/purposes.njk
```diff
@@ -0,0 +1,113 @@
...
+          </details>
+        {% endif %}
+      </dd>
+    </div>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key govuk-!-font-weight-regular">
+        Period of abstraction
+      </dt>
+      <dd class="govuk-summary-list__value">
+        From {{ purpose.periodStart | abstractionPeriod() }}
```
james-siteclick: If the filter has no arguments, I think the brackets may be optional, e.g.:
`{{ purpose.periodStart | abstractionPeriod }}`

src/external/lib/view-engine/filters/ngr-points.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Creates extraction point/s description string based on number of grid reference points
+ * @param  {Object} points contains details of extraction point, including 1, 2 or 4 grid references
+ * @return {String}        description of extraction points
+ */
+const ngrPoints = points => {
```
james-siteclick: This feels like it could be neatly moved to a Nunjucks macro to display a single NGR because it is presentational.
I think we probably only need a filter for `formatGridReference`.

src/external/lib/view-engine/filters/ngr-points.js
```diff
@@ -0,0 +1,47 @@
...
+    response += ` (${point.name})`;
+  }
+
+  return response;
+};
+
+/**
+ * Format NGR point string, e.g. ST123456 so it has spaces, e.g. ST 123 456
+ */
+const formatGridReference = reference => {
```
james-siteclick: Could we add a test for this function?

src/external/lib/view-engine/filters/ngr-points.js
DGajewska: I realised that this whole filter was a duplicate of a macro that was already created, so I will be deleting it and using the macro instead

src/external/views/nunjucks/view-licences/purposes.njk
DGajewska: Yes, loop.last works. I wasn't aware of its existence. Now updated

PR #783, opened 2019-05-24
Chore/move shared plugins

swaff:


src/shared/plugins/csrf/index.js
```diff
@@ -31,43 +31,43 @@ function validatePayload (payload) {
...
+    }
+  }
+};
 
-        // Check referrer if header is set
-        if (request.headers.referer) {
-          const currentHost = new URL(`${request.info.protocol}://${request.info.host}`);
-          const refererUrl = new URL(request.headers.referer);
+const validateCsrfToken = request => {
+  const token = request.sessionStore.get('csrf_token');
```
james-siteclick: I think this needs to become:
```
request.yar.get('csrfToken')
```

test/shared/lib/plugings/config/index.js
```diff
@@ -0,0 +1,48 @@
+const hapi = require('@hapi/hapi');
```
james-siteclick: Looks like we may have some `plugings`

PR #800, opened 2019-05-31
Create shared password reset plugin

james-siteclick:


test/shared/plugins/reset-password/controller.js
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * Note current version test of 'code' can't test for async function
+ * @see {@link https://github.com/hapijs/code/issues/103}
+ */
+function isAsync (fn) {
+  return fn.constructor.name === 'AsyncFunction';
+}
+
+experiment('reset password controller', () => {
+  experiment('Check methods on reset password controller', () => {
```
swaff: I don't think we need these tests as the tests are exercised in the experiments below.

PR #805, opened 2019-06-04
Feat: Update view-licences/conditions page to Nunjucks

DGajewska:
WATER-2145

Update routes and helpers files. Create conditions page and licence-conditions macro. Delete html conditions page. Minor tweaks to purposes njk page for consistency.


src/internal/views/nunjucks/view-licences/macros/licence-conditions.njk
```diff
@@ -0,0 +1,102 @@
...
+          </dd>
+        </div>
+        {% endif %}
+
+
+        {% endfor %}
+      </dl>
+    </div>
+  </div>
+  {% if loop.index != loop.last %}<br>{% endif %}
```
james-siteclick: I think `loop.last` is a boolean not an integer.  So maybe this should just be:
```
{% if loop.last %}<br>{% endif %}
```
or even the inline form:
```
{{ '<br>' if loop.last }}
```

DGajewska: I changed it to `{% if not loop.last %}<br>{% endif %}` which is what I wanted, just wasn't sure how to do it

PR #806, opened 2019-06-04
Chore: Moves the notifications reports to nunjcucks

swaff:


src/shared/assets/sass/v2/_utils.scss
```diff
@@ -1,3 +1,20 @@
 .borderless {
   border: none;
 }
+
+// Screen-reader only text
+// @see https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/
+.sr-only {
```
james-siteclick: We could use the standard class instead of `sr-only`:
```
govuk-visually-hidden
```

PR #817, opened 2019-06-07
Chore/1792 move shared lib code to shared dir

swaff:


src/external/lib/view-engine/filters/index.js
```diff
@@ -1,19 +1,17 @@
...
-  ...require('./number'),
-  ...require('./pluralize'),
-  ...require('./query-string'),
-  ...require('./slice'),
-  ...require('./sort-new-direction'),
-  ...require('./sort-query'),
-  ...require('./title-case'),
-  ...require('./unit-conversion'),
-  ...require('./units')
+  ...require('../../../../shared/view/nunjucks/filters/abstraction-period'),
```
james-siteclick: Would these work without the `../../../../` since we are using the `app-module-path` node module?

src/external/modules/returns/lib/return-path.js
```diff
@@ -37,7 +37,7 @@ const isInternalEdit = (ret, request) => {
 const getEditButtonPath = (ret, request) => {
   if (isInternalEdit(ret, request)) {
     const returnId = getReturnId(ret);
-    return `/admin/return/internal?returnId=${returnId}`;
+    return `/return/internal?returnId=${returnId}`;
```
james-siteclick: I think this whole function may need removing from the external app - as external users shouldn't get an edit link.

src/external/modules/view-licences/controller.js
```diff
@@ -158,7 +158,7 @@ const getCommonLicenceViewContext = async (licenceNumber, documentId, documentNa
     documentId,
     ...returnsData,
     pageTitle: getPageTitle(documentName, licenceNumber),
-    back: `/admin/licences?query=${licenceNumber}`,
+    back: `/licences?query=${licenceNumber}`,
```
james-siteclick: Unless it's been changed elsewhere, the `?query=` is not a valid query param for the external licences view and causes a 500 error.  For externals there are 2 query params `emailAddress` and `licenceNumber`

src/external/views/nunjucks/returns/macros/return-header.njk
```diff
@@ -78,7 +78,7 @@
     </dl>
 
     <p class="medium-space">
-      <a href="{%if isAdmin %}/admin{%endif%}/{{ "expired-" if not documentHeader.metadata.IsCurrent }}licences/{{ documentHeader.document_id }}">View this licence</a>
+      <a href="/{{ "expired-" if not documentHeader.metadata.IsCurrent }}licences/{{ documentHeader.document_id }}">View this licence</a>
```
james-siteclick: I don't think view of expired licence is supported for external user

src/external/views/nunjucks/view-licences/expired-licence.njk
```diff
@@ -19,7 +19,7 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
       {% if primaryUser %}
```
james-siteclick: I think this whole if block can be removed for external

src/external/views/nunjucks/view-licences/licence.njk
```diff
@@ -15,7 +15,7 @@
       {% endif %}
 
       {% if isInternal and primaryUser %}
```
james-siteclick: I think this whole if block can be removed for external

src/internal/modules/returns/controllers/upload.js
```diff
@@ -5,13 +5,13 @@ const snakeCase = require('snake-case');
```
james-siteclick: Internal users can't upload, so this controller could be removed

src/internal/modules/returns/lib/csv-templates.js
```diff
@@ -7,7 +7,7 @@ const util = require('util');
 const csvStringify = util.promisify(require('csv-stringify'));
```
james-siteclick: This could be removed for internals

PR #822, opened 2019-06-11
Fix: Change back link text to 'Licence number xxx/xxx'

DGajewska:
WATER-2145


src/external/modules/view-licences/controller.js
```diff
@@ -61,7 +61,8 @@ async function getLicenceDetail (request, reply) {
       licence_id: documentId,
       name: 'name' in request.view ? request.view.name : customName,
       licenceData: viewData,
-      back: `/licences/${documentId}`,
+      back: isInternal(request) ? `/admin/licences/${documentId}` : `/licences/${documentId}`,
```
james-siteclick: I think the isInternal() logic should be removed this sprint, the removed line in red is correct

src/external/views/nunjucks/layout.njk
```diff
@@ -81,10 +81,16 @@
   {% endblock %}
 
   {% if back %}
+
+  {% set backLinkText %}
+    {% if backText %} {{ backText }} {% else %} Back {% endif %}
```
james-siteclick: This could be written more compactly:
```
{{ backText | default('Back') }}
```

src/internal/views/nunjucks/layout.njk
```diff
@@ -81,10 +81,16 @@
   {% endblock %}
 
   {% if back %}
+
+  {% set backLinkText %}
+    {% if backText %} {{ backText }} {% else %} Back {% endif %}
```
james-siteclick:
```
{{ backText | default('Back') }}
```

PR #823, opened 2019-06-11
Update zip folder and file names for csv returns

DGajewska:
WATER-2143

Change file & zip names from snakeCase to lowerCase. Include 'return/s' in file name. Include 'return templates' and year in zip file name.


swaff:
This can be merged if the zip download has been tested. I was just wondering if the filename needs quotes due to the spaces?

PR #831, opened 2019-06-13
Chore: Move the data access to shared location

swaff:
WATER-1791

 - Moves the document-verifications connector to the shared services dir.
 - Replaces documents connector with shared service
 - Moves entities connectors to EntitiesAPIClient
 - Moves the CRM KPI connector to shared services
 - Renames API clients to use ApiClient over APIClient
 - Moves the crm verifications connector to the VerificationsApiClient
 - Replaces any use of connectors/crm with `services.crm`
 - Adds the app-module-path to labrc so the tests can use the simplified
require path.
 - Propagates use of app-module-path for shared content


src/external/lib/connectors/crm.js
```diff
@@ -162,7 +157,7 @@ async function verify (entityId, verificationCode) {
   }
 
   // Verify with code
```
james-siteclick: I have added a Trello card for moving the registration logic to the water service

PR #838, opened 2019-06-19
Feat/water 2148

DGajewska:
Create gauging-station Nunjucks page. 
Update gauging station controller and route functions to use load licenceData plugin. 
Create helper function to add 'isHof' flag for each condition in the licence data. 
Update helpers to set pageTite and pageHeading for each page. 
Update conditions, contact, points and purposes pages for consistency - pageHeadings and licence-condition macro. 
Create macros gauging-station-measurement and point-string. 
Create Nunjucks filters: flow-converter and gauging-station-value.


src/external/lib/view-engine/filters/index.js
```diff
@@ -2,7 +2,9 @@ module.exports = {
   ...require('../../../../shared/view/nunjucks/filters/abstraction-period'),
```
james-siteclick: We are using a module `app-module-path` which is supposed to resolve the `/src` directory.

So you may find you can rewrite these as e.g.
`require('shared/view/nunjucks/filters/abstraction-period')`

src/external/modules/view-licences/controller.js
```diff
@@ -121,38 +124,58 @@ async function postLicenceRename (request, h) {
...
- * Displays a gauging station flow/level data, along with HoF conditions
- * for the selected licence
+ * Throws a 404 error if the gauging station requested is not attached to the
+ * licence referred to in the document ID param
+ * @param  {Object} request - HAPI request
  */
-async function getLicenceGaugingStation (request, reply) {
-  const { measure: mode } = request.query;
-  const { licence_id: documentHeaderId, gauging_station: gaugingStation } = request.params;
+const validateGaugingStation = request => {
```
james-siteclick: Can we add a test for this

src/external/modules/view-licences/controller.js
```diff
@@ -121,38 +124,58 @@ async function postLicenceRename (request, h) {
...
-  if (!helpers.validateStationReference(licenceData.permitData.metadata.gaugingStations, gaugingStation)) {
-    throw Boom.notFound(`Gauging station ${gaugingStation} not linked to licence ${licenceData.documentHeader.system_external_id}`);
-  }
+const isFlowCondition = condition =>
+  (condition.code === 'CES' && condition.subCode === 'FLOW');
 
-  // Load river level data
-  const { hofTypes } = licenceData.viewData;
-  const { riverLevel, measure } = await helpers.loadRiverLevelData(gaugingStation, hofTypes, mode);
+const getHoFTypes = (conditions = []) => ({
```
james-siteclick: Could we add a test for this

src/external/modules/view-licences/helpers.js
```diff
@@ -262,6 +270,26 @@ function errorMapper (error) {
...
+ */
+const setConditionHofFlags = (viewContext) => {
+  const updated = cloneDeep(viewContext);
+
+  const conditions = updated.summary.conditions;
+
+  const updatedConditions = conditions.map(condition => {
+    return {
+      ...condition,
+      isHof: (condition.code === 'CES' && (condition.subCode === 'FLOW' || condition.subCode === 'LEV'))
```
james-siteclick: This might be a bit shorter, not sure if it's better:
```
['FLOW', 'LEV'].includes(condition.subCode)
```

src/external/views/nunjucks/view-licences/macros/point-string.njk
```diff
@@ -0,0 +1,3 @@
+{% macro pointString(point) %}
```
james-siteclick: I assume the NGRs are always 8 char?

DGajewska: I just had a look at the handlebars filter that I was recreating and realised that it's done differently there, will update it to match.

PR #858, opened 2019-06-28
WATER-1800

swaff:
- Creates service status plugin to show service status at
/service-status in external and internal applications.

- Moves core logic to water-service

- Uses GDS styles for HTML version of the service status page

- Removes idm and crm KpiApiClient classes which are redundant.


src/external/views/nunjucks/service-status/index.njk
```diff
@@ -1,68 +1,67 @@
-<html>
-   <head>
-   <body>
+{% extends "./nunjucks/layout.njk" %}
```
james-siteclick: Should we now hide all the stats for external users and just show 'OK'?

swaff: I think we agreed that this desicion would be outside of this story.

PR #874, opened 2019-07-05
Feat: Change email UI

swaff:
WATER-2154

Update email UI and company selector update

 - Includes new account-setting landing page
   - Update email link commented out at this stage
 - Includes confirm password page
 - Adds page for adding new email address
 - Adds page for email verification code
 - Adds success page
 - Adds the try again page for failed password attempts

Reworks the logic for the company selector which can now use the
`request.defra` object to help determine if a user should be redirected,
and also adds a settings item to allow a route to opt out of company
selector redirection. This came about because the user is logged in when
updating their account details so the existing mechanism of checking if
the route did not require authentication was no longer relevant.

### DIFF
--- a/src/external/lib/hapi-plugins/company-selection.js
+++ b/src/external/lib/hapi-plugins/company-selection.js
@@ -7,18 +7,22 @@
  * add-licences route.
  *
  * Both of the above redirections will only take place if the user is
- * attempting to access a route that is configured with an access scope.
+ * attempting to access a route that has not opted out by setting
+ * config.plugins.companySelector.ignore to true.
  */
-const { get, isEmpty } = require('lodash');
+const { get } = require('lodash');
 const SELECT_COMPANY_PATH = '/select-company';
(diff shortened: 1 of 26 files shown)


src/external/lib/hapi-plugins/company-selection.js
```diff
@@ -7,26 +7,30 @@
...
 const ADD_LICENCES_PATH = '/add-licences';
 
 const shouldRedirect = request => {
-  const { companyId } = request.defra;
-  const { path } = request;
-  const access = get(request, 'route.settings.auth.access', {});
+  if (request.method.toLowerCase() === 'get' && request.defra) {
+    const { companyId } = request.defra;
+    const { path } = request;
+    const ignoreRoute = get(request, 'route.settings.plugins.companySelector.ignore', false);
```
james-siteclick: I wonder if we could use scopes for this?
E.g. if a route requires authentication but not a specific scope, does this indicate that it's a route which does not require company selection?

PR #887, opened 2019-07-15
Feat: Adds user journey info to error reporting

swaff:
WATER-2166

Adds a shared plugin that records the request made by a user to session
to help with bug fixing when errors are logged.

Changes the logger setup so that the logger is monkey patched with a
method for logging errors with user journeys.

Uses the new logger method in all places where the HAPI Request object
is available.

Updates dependencies to handle some security advisories from npm.


src/external/lib/hapi-plugins/error.js
```diff
@@ -46,7 +46,7 @@ const _handler = async (request, h) => {
     return h.redirect('/welcome');
   }
 
-  logger.error(pick(res, ['error', 'message', 'statusCode', 'stack']));
+  logger.error('Unexpected error', pick(res, ['error', 'message', 'statusCode', 'stack']));
```
james-siteclick: Do we want to include the `errorWithJourney` here?

PR #909, opened 2019-07-30
Move returns index page to Nunjucks.

DGajewska:
WATER-2040

Create Nunjucks returns index page. Delete html return index pages.


src/external/modules/returns/controllers/view.js
```diff
@@ -56,7 +53,7 @@ const getReturn = async (request, h) => {
   const { licenceNumber } = data;
 
   // Load licence from CRM to check user has access
-  const [ documentHeader ] = await getLicenceNumbers(request, { system_external_id: licenceNumber, includeExpired: false });
+  const [ documentHeader ] = await helpers.getLicenceNumbers(request, { system_external_id: licenceNumber, includeExpired: false });
 
   const canView = documentHeader && data.isCurrent && model.metadata.isCurrent;
```
james-siteclick: Where does `model` come from here?

src/external/views/nunjucks/returns/index.njk
```diff
@@ -0,0 +1,102 @@
...
+                      {{ return.licenceName if return.licenceName else return.licence_ref}}
+                      {% if return.path %}
+                        <span class="govuk-visually-hidden">
+                          for
+                          {% for purpose in return.metadata.purposes %}
+                            {{ purpose.tertiary.description }}
+                          {% endfor %}
+                          at
+                          {{ return.metadata.description | titleCase }}
+                          {% if return.received_date == null %}
```
james-siteclick: I wonder if we should be checking `return.status==='due'`

src/external/views/nunjucks/returns/index.njk
```diff
@@ -0,0 +1,102 @@
...
+                  </td>
+                  <td class="govuk-table__cell table__cell">
+                    <h4 class="govuk-visually-hidden">Purpose</h4>
+                    {% for purpose in return.metadata.purposes %}
+                      {{ purpose.alias | titleCase if purpose.alias else purpose.tertiary.description | titleCase }}
+                    {% endfor %}
+                  </td>
+                  <td class="govuk-table__cell table__cell">
+                    <h4 class="govuk-visually-hidden">Status</h4>
+                    <div class="badge badge--{{ (return | returnBadge).status }} badge--no-wrap">
```
james-siteclick: Do we have a badge macro?

src/shared/assets/sass/v2/blocks/_table.scss
```diff
@@ -46,33 +13,93 @@
...
+  // Hint to user this is possible by displaying hover state on table row
+  .js-enabled & {
+    &:hover {
+      cursor: pointer;
+      background: #f3f2f1;
+    }
   }
 }
 
+.column--15 { width: 15% }
```
james-siteclick: Can we check if there are standard gov-uk classes that would help with this and the class below?
We should always avoid adding custom CSS wherever possible

src/shared/assets/sass/v2/blocks/_table.scss
```diff
@@ -46,33 +13,93 @@
...
 
+.big-space {
+  margin-bottom: 45px;
+}
 
+// .table {
```
james-siteclick: Let's delete all this commented stuff if it's no longer needed

PR #911, opened 2019-07-31
Feat: Adds the UI for adding an internal user

swaff:
WATER-2227

Adds the UI component of the larger feature to allow an internal user to
create other internal users and assign a set of permissions.

### DIFF
--- a/src/internal/modules/account/forms/set-permissions.js
+++ b/src/internal/modules/account/forms/set-permissions.js
@@ -1,0 +1,79 @@
+const Joi = require('@hapi/joi');
+const { formFactory, fields, setValues } = require('shared/lib/forms');
+
+const getChoices = () => ([
+  {
+    value: 'basic',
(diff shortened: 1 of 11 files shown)


src/internal/modules/account/forms/set-permissions.js
```diff
@@ -0,0 +1,79 @@
...
+
+  f.fields.push(fields.hidden('csrf_token', {}, csrfToken));
+  f.fields.push(fields.button(null, { label: 'Continue' }));
+
+  return setValues(f, { permission });
+};
+
+const schema = {
+  csrf_token: Joi.string().uuid().required(),
+  permission: Joi.string().required()
```
james-siteclick: Should we limit `permission` to the available values in the choices array above? e.g.
```
const getChoiceValues = field => field.options.choices.map(choice => choice.value);
```

PR #914, opened 2019-08-01
Create new 'manage' tab in Nunjucks and remap roles/permissions/scope…

james-siteclick:
- Implements the roles returned from the updated IDM API
- Replaces the Notifications & Reports tab with a new Manage tab 
- Updates scopes to reflect the new roles


test/internal/lib/permissions.js
```diff
@@ -21,14 +21,33 @@ const createRequest = (scope) => ({
...
-    test('it should return true if scope is in the credentials', async () => {
-      const request = createRequest(['foo', 'bar']);
-      expect(permissions.hasScope(request, 'foo')).to.equal(true);
+    experiment('when the supplied scope is a string', () => {
+      test('it should return true if scope is in the credentials', async () => {
+        const request = createRequest(['foo', 'bar']);
+        expect(permissions.hasScope(request, 'foo')).to.equal(true);
+      });
+
+      test('it should return true if scope is not in the credentials', async () => {
```
swaff: it should return false if scope is not in the credentials

PR #919, opened 2019-08-06
Chore: Create Nunjucks page for return page.

DGajewska:
WATER-2040

Create return Nunjucks page. Delete return html pages. Move all shared return macros into shared/. Implement clickable functionality on previous versions. Add test coverage.


src/external/views/nunjucks/returns/meter-reset.njk
```diff
@@ -1,10 +1,10 @@
...
 {% from "forms/index.njk" import formRender, formErrorSummary %}
-{% from "nunjucks/returns/macros/return-header.njk" import returnHeader %}
+{% from "returns/return-header.njk" import returnHeader %}
 
 {% block content %}
   {{ formErrorSummary(form) }}
-  {{ returnHeader(data, documentHeader, showMeta, isAdmin) }}
+  {{ returnHeader(data, documentHeader, showMeta) }}
   {{ formRender(form) }}
 {% endblock %}
```
james-siteclick: I think we should probably still use `var` instead of `const` in front-end JS.

Is there any way this functionality could be generalised and placed in the main application JS?  E.g. in the same way the clickable tables was implemented with the data attribute.

This template would then no longer be necessary.

src/external/views/nunjucks/returns/return.njk
```diff
@@ -0,0 +1,57 @@
...
+      </div>
+    {% else %}
+      {% if data.isNil %}
+        <h2 class="govuk-heading-l">Nil return</h2>
+      {% else %}
+        <h2 class="govuk-heading-m">Abstraction volumes</h2>
+      {% endif %}
+    </div>
+
+    <div class="govuk-grid-column-full">
```
james-siteclick: I think there is a standard contact details macro somewhere we could use with email/phone etc.

src/shared/assets/sass/v2/blocks/_inset-text.scss
```diff
@@ -0,0 +1,14 @@
...
+
+  border-color: $govuk-link-colour;
+}
+
+.inset-text--clickable {
+  @extend .govuk-inset-text;
+
+  &:hover {
+    cursor: pointer;
+    background: #f3f2f1;
```
james-siteclick: I think we can use:
```
 govuk-colour("light-grey")
```

PR #933, opened 2019-08-14
Finish work on 'Create new internal users' feature

DGajewska:
WATER-2182

Update controller to work with 1 call to the water service to create user with permissions. Move connector functions to helpers file. Add serviceRequest to create internal user.


src/internal/modules/account/controller.js
```diff
@@ -56,27 +52,27 @@ const getSetPermissions = async (request, h, formFromPost) => {
...
     setPermissionsForm(request, payload),
     request,
     setPermissionsSchema
   );
 
   if (form.isValid) {
-    // TODO: Update the users permissions
-
-    // then redirect to the success page
+    const newUser = await helpers.getInternalUser(callingUserId, newUserEmail, permission);
```
james-siteclick: I think if the user has been successfully created here, this is where the session key should be deleted.
Should we be using:
```
delete request.yar.clear('key') 
```
as per the docs?

src/internal/modules/account/helpers.js
```diff
@@ -0,0 +1,33 @@
...
+    : /\.gov\.uk$/;
+};
+
+const getUserById = async (userId) => {
+  const { data: user, error } = await services.idm.users.findOne(userId);
+  throwIfError(error);
+  return user;
+};
+
+const getUserByEmail = async (email, config = internalConfig) => {
```
james-siteclick: I think there may already be a `findOneByEmail` method on `services.idm.users` connector

src/internal/modules/account/helpers.js
```diff
@@ -0,0 +1,33 @@
...
+const internalConfig = require('internal/config');
+const services = require('internal/lib/connectors/services');
+
+const getEmailRegex = (config = internalConfig) => {
+  return (config.isLocal || config.testMode)
+    ? /(\.gov\.uk|gmail\.com)$/
+    : /\.gov\.uk$/;
+};
+
+const getUserById = async (userId) => {
```
james-siteclick: This looks like a useful function, maybe we should add this to the connector at `services.idm.users.findOneById` so it can be used elsewhere

src/internal/modules/account/helpers.js
```diff
@@ -0,0 +1,33 @@
...
+const getUserByEmail = async (email, config = internalConfig) => {
+  const userData = await services.idm.users.findOneByEmail(email, config.idm.application);
+  if (userData) {
+    const { data: user, error } = userData;
+    throwIfError(error);
+    return user;
+  }
+};
+
+const getInternalUser = async (callingUserId, newUserEmail, permission) => {
```
james-siteclick: This function doesn't appear to do anything.  Could the service function be used directly in the controller?

src/internal/modules/account/helpers.js
```diff
@@ -0,0 +1,33 @@
+const { throwIfError } = require('@envage/hapi-pg-rest-api');
+const internalConfig = require('internal/config');
+const services = require('internal/lib/connectors/services');
+
+const getEmailRegex = (config = internalConfig) => {
```
james-siteclick: If this ends up in a file all by itself it could move back to the form, as I think it is only used there

PR #940, opened 2019-08-19
Feat: Amend internal user permissions

DGajewska:
WATER-2183

Add permissions form to user-status page. Create update-permissions-success.njk page. Add connector to water service endpoint. Add routes and controller functions for post and get requests.

### DIFF
--- a/src/internal/modules/account/controller.js
+++ b/src/internal/modules/account/controller.js
@@ -45,7 +45,7 @@ const postCreateAccount = async (request, h) => {
 };
 
 const getSetPermissions = async (request, h, formFromPost) => {
-  const form = formFromPost || setPermissionsForm(request);
+  const form = formFromPost || setPermissionsForm(request, '/account/create-user/set-permissions');
 
   return h.view(
     'nunjucks/account/set-permissions.njk',
@@ -61,7 +61,7 @@ const postSetPermissions = async (request, h) => {
   const { userId: callingUserId } = request.defra;
   const { newUserEmail, permission } = request.payload;
   const form = handleRequest(
-    setPermissionsForm(request, request.payload),
+    setPermissionsForm(request, request.payload, true),
     request,
     setPermissionsSchema
   );
@@ -72,7 +72,8 @@ const postSetPermissions = async (request, h) => {
 
   try {
     const newUser = await services.water.users.postCreateInternalUser(callingUserId, newUserEmail, permission);
-    delete request.yar.clear('key');
+    request.yar.clear('newInternalUserAccountEmail');
+
     return h.redirect(`/account/create-user/${newUser.user_id}/success`);
   } catch (err) {
     // User exists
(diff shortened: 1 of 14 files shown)


src/internal/modules/account/controller.js
```diff
@@ -45,7 +45,7 @@ const postCreateAccount = async (request, h) => {
 };
 
 const getSetPermissions = async (request, h, formFromPost) => {
-  const form = formFromPost || setPermissionsForm(request);
+  const form = formFromPost || setPermissionsForm(request, '/account/create-user/set-permissions');
```
james-siteclick: Here the second argument to`setPermissionsForm` is a URL path 
In the call below it is the request payload. 
Is this correct?

src/internal/modules/account/forms/set-permissions.js
```diff
@@ -65,8 +67,10 @@ const form = (request, permission) => {
     }
   }));
 
-  const newAccountEmail = get(request, 'yar._store.newInternalUserAccountEmail');
-  f.fields.push(fields.hidden('newUserEmail', {}, newAccountEmail));
+  if (newUser) {
+    const newAccountEmail = get(request, 'yar._store.newInternalUserAccountEmail');
```
james-siteclick: The _ often denotes a variable that should be considered private.  Should we be using the documented API, e.g.:
```
request.yar.get('newInternalUserAccountEmail');
```

src/internal/modules/internal-search/controller.js
```diff
@@ -58,19 +60,74 @@ const getOutstandingVerificationLicenceCount = companies => {
...
-  const { view } = request;
+const getPermissionsFormData = async (request) => {
+  if (isManageAccounts(request)) {
+    const user = await services.idm.users.findOneById(request.params.userId);
+    const permission = get(user, 'groups[0]');
+    return setPermissionsForm(request, permission);
+  }
+};
+
+const getPermissionsLabelText = user => {
```
james-siteclick: Maybe something like:
```
const group = groups[0] || 'basic';
const role = roles[0];

return [group, role].filter(x => x).join('_');
```

src/internal/modules/internal-search/routes.js
```diff
@@ -41,5 +41,47 @@ module.exports = {
...
+      auth: { scope: 'manage_accounts' },
+      plugins: {
+        viewContext: {
+          pageTitle: 'User status',
+          activeNavLink: 'view'
+        }
+      },
+      validate: {
+        params: {
+          userId: Joi.number()
```
james-siteclick: Maybe we should add `.required()` to all the user ID validators

src/internal/modules/internal-search/controller.js
```diff
@@ -58,19 +60,74 @@ const getOutstandingVerificationLicenceCount = companies => {
...
+    const user = await services.idm.users.findOneById(request.params.userId);
+    const permission = getPermissionsKey(user);
+    return setPermissionsForm(request, permission);
+  }
+};
+
+const getPermissionsKey = user => {
+  const group = user.groups[0] || 'basic';
+  const roles = user.roles.filter(role => role.startsWith('ar_'));
+  console.log('permissions key', [group, ...roles].join('_'));
```
james-siteclick: Remove `console.log`

PR #943, opened 2019-08-20
Bugfix/water 2268 contact details

james-siteclick:


src/internal/modules/notifications/forms/name-and-job.js
```diff
@@ -0,0 +1,40 @@
+const Joi = require('@hapi/joi');
+const { formFactory, fields, setValues } = require('shared/lib/forms');
+
+const getTextField = (name, label, multiline = false) =>
+  fields.text(name, {
+    label,
+    multiline,
```
DGajewska: This line is redundant, is it worth removing?

PR #963, opened 2019-09-04
Feat: Unlink licence

DGajewska:
WATER-2238

Add views, forms, routes and controller for unlink licence. Add company data call to licence plugin.


src/internal/views/nunjucks/internal-search/macros/company-licences.njk
```diff
@@ -20,6 +21,13 @@
               <a class="govuk-link" href="/licences/{{licence.documentId}}">{{licence.licenceRef}}</a>
             </td>
             <td class="govuk-table__cell">{{licence.licenceHolder}}</td>
+
+            <td class="govuk-table__cell govuk-table__cell--numeric">
+              {% if userCanUnlinkLicences %}
+                <a class="govuk-link" href="/licences/{{licence.documentId}}/{{unlinkLicencePathTail}}">Unlink licence</a>
```
james-siteclick: For accessibility, we may want something like:
```
Unlink licence<span class="govuk-visually-hidden"> {{ licence.licenceNumber }}</span>
```
This avoids having multiple links on the page with the same text but pointing to a different page

PR #974, opened 2019-09-06
Chore: Removes internal and external logic now apps are split

swaff:
WATER-1797

- Removes internal specific logic from external application
- Removes external specific logic from internal application
- Removes redundant folder for view-licences__orig
- Moves repeated returns helper code to shared/lib/returns


src/shared/lib/returns/badge.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Gets badge object to render for return row
+ * @param  {String}  status    - return status
+ * @param  {Boolean} isPastDue - whether return is past due
+ * @return {Object}            - badge text and style
+ */
+const getBadge = (status, isPastDueDate) => {
+  const viewStatus = ((status === 'due') && isPastDueDate) ? 'overdue' : status;
+
+  const styles = {
```
james-siteclick: I have recently updated the badge classes in another PR:
https://github.com/DEFRA/water-abstraction-ui/pull/975

PR #995, opened 2019-09-23
Pass isFinal flag in metadata to getRequiredLines helper function.

DGajewska:
WATER-2012

Pass isFinal flag in csv-templates, WaterReturn and Line models.


src/shared/modules/returns/models/WaterReturn.js
```diff
@@ -29,7 +29,7 @@ class WaterReturn {
     this.isNil = data.isNil;
 
     const lineOptions = pick(data, ['startDate', 'endDate', 'frequency']);
-    this.lines = new Lines(data.lines, lineOptions);
```
james-siteclick: I wonder if we will need to do:
```
const lineOptions = {
  ...pick(data, ['startDate', 'endDate', 'frequency']),
  isFinal : get(data, 'metadata.isFinal', false)
}
```
The reason is that the flag won't be present in the metadata for up to 48 hours following deployment (when the import process does a full data refresh).
In this situation I think the Lines might crash because it is expecting boolean for this flag but it will get undefined.
Another alternative could be to implement this default value in the water service.

PR #1004, opened 2019-10-03
Feat: Facilitate acceptance tests

swaff:
WATER-2230

Provides a proxy through to the /acceptance-tests routes in the water
service to allow the acceptance tests to set up and tear down data.

### DIFF
--- a/src/shared/plugins/acceptance-tests-proxy/controller.js
+++ b/src/shared/plugins/acceptance-tests-proxy/controller.js
@@ -1,0 +1,18 @@
+const Boom = require('boom');
+
+/**
+ * Takes the path after "acceptance-tests" and forwards the
+ * request to the water-service to faciliate acceptance testing.
+ *
+ * @param {Object} request HAPI request
+ */
+const proxyToWaterService = request => {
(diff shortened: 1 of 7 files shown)


src/shared/plugins/acceptance-tests-proxy/controller.js
```diff
@@ -0,0 +1,18 @@
+const Boom = require('boom');
+
+/**
+ * Takes the path after "acceptance-tests" and forwards the
+ * request to the water-service to faciliate acceptance testing.
+ *
+ * @param {Object} request HAPI request
+ */
+const proxyToWaterService = request => {
+  if (process.env.NODE_ENV === 'production') {
```
james-siteclick: I think for safety this should whitelist allowed environments

src/shared/plugins/acceptance-tests-proxy/routes.js
```diff
@@ -0,0 +1,13 @@
+const controller = require('./controller');
+
+if (process.env.NODE_ENV !== 'production') {
```
james-siteclick: As above if we could whitelist allowed environments

PR #1014, opened 2019-10-10
Internal Search page accessibility bug fix

Stephan-Hough:
WATER-2327
- Changed the hint to a label on internal search form. This changes the colour of the text but was approved by UX Design.

### DIFF
--- a/src/internal/modules/internal-search/forms/search-form.js
+++ b/src/internal/modules/internal-search/forms/search-form.js
@@ -12,7 +12,7 @@ const form = (query) => {
 
   f.fields.push(fields.text('query', {
     widget: 'search',
-    hint: 'Search by name, registered email address, return ID or licence number',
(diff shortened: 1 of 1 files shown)


src/internal/modules/internal-search/forms/search-form.js
```diff
@@ -12,7 +12,7 @@ const form = (query) => {
 
   f.fields.push(fields.text('query', {
     widget: 'search',
-    hint: 'Search by name, registered email address, return ID or licence number',
+    label: 'Search by name, registered email address, return ID or licence number',
```
james-siteclick: The above solution would work, although it would change the colour of the text.  It looks like this page already has a label implemented in the H1 tag, however the input field it references does not have an id attribute set to `query`.

So it may be worth trying:
```
  id : 'query`
```
or 
```
  attr : { id : 'query' }
```

PR #1016, opened 2019-10-10
Accessibility bug - changed CSS styling to increase contrast

Stephan-Hough:
WATER-2326
- created a new css class for the rename licence box.
- added the styling to the inset-text.scss file - (not sure if this is the right place to add)
- removed well.scss as this is no longer used anywhere else.


src/shared/assets/sass/v2/blocks/_inset-text.scss
```diff
@@ -12,3 +12,10 @@
     background: govuk-colour("grey-4");
   }
 }
+
+.inset-text--rename-licence {
+  @extend .govuk-inset-text;
+  border: 2px solid $govuk-text-colour;
```
james-siteclick: I wonder if for the border we should use: 
```
border: 1px solid $govuk-border-colour;
```
As this would match more closely some of the other components in the kit?

src/internal/views/nunjucks/view-licences/rename.njk
```diff
@@ -9,7 +9,7 @@
 
   {{ title('Licence number ' + summary.licenceNumber, summary.documentName ) }}
 
-  <div class="well">
+  <div class="inset-text--rename-licence">
```
james-siteclick: Hopefully you should find the `rename` feature has been removed from the internal user site, so it should be OK to delete this template.

src/shared/assets/sass/v2/blocks/_inset-text.scss
Stephan-Hough: I agree. Will get check with UX design again if this is how they want to proceed.

Stephan-Hough: Form has not been redesigned in the last commit

src/internal/views/nunjucks/view-licences/rename.njk
Stephan-Hough: File now deleted

PR #1018, opened 2019-10-11
Fix: Makes the no data tabs more consistent

swaff:
WATER-2245

When displaying licence data, only show returns/charges/communications
tabs if there is data to show.


src/internal/views/nunjucks/view-licences/expired-licence.njk
```diff
@@ -22,43 +22,48 @@
...
 
-      {% set returnsHtml %}
-        {% include "nunjucks/view-licences/tabs/returns.njk" %}
-      {% endset %}
+      {% if (returns.length > 0 or communications.length > 0 or (showChargeVersions and chargeVersions.length > 0)) %}
+        <div class="govuk-tabs" data-module="tabs">
+          <h2 class="govuk-tabs__title">Contents</h2>
+          <ul class="govuk-tabs__list">
+            {% if returns.length > 0 %}
+              <li class="govuk-tabs__list-item">
```
james-siteclick: Does the govuk-tabs component automatically select the first link in the tab navigation?
The HTML for the component shows `govuk-tabs__list-item--selected` as the class on the first element

src/internal/views/nunjucks/view-licences/licence.njk
```diff
@@ -28,15 +29,17 @@
             </a>
           </li>
 
-          {% if returns.length %}
+          {% if returns.length > 0 %}
             <li class="govuk-tabs__list-item">
```
james-siteclick: Same here, should the first LI have `govuk-tabs__list-item--selected` or is it preselected without it?

src/internal/views/nunjucks/view-licences/expired-licence.njk
swaff: The govuk component does set the selected tab.

src/internal/views/nunjucks/view-licences/licence.njk
swaff: In this case the summary tab is always the selected tab.

PR #1050, opened 2019-11-07
Update contactDetails Macro & Accessibility Page.

DGajewska:
WATER-2325

Update contactDetails macro to allow for different email addresses to be used & for the option of displaying it single spaced.

### DIFF
--- a/src/shared/view/nunjucks/macros/contact-details.njk
+++ b/src/shared/view/nunjucks/macros/contact-details.njk
@@ -1,13 +1,27 @@
-{% macro contactDetails() %}
+{% macro contactDetails(emailAddress, isSingleSpaced) %}
 
-  <p>Email: {{ 'enquiries@environment-agency.gov.uk' | urlize | safe }}</p>
+  {% set openingPTag%}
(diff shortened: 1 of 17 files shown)


src/shared/view/nunjucks/macros/contact-details.njk
```diff
@@ -1,13 +1,27 @@
-{% macro contactDetails() %}
+{% macro contactDetails(emailAddress, isSingleSpaced) %}
```
swaff: I think it might be nicer if the signature for this macro is something like

```JavaScript
{% macro contactDetails(options = {
  email: 'enquiries@environment-agency.gov.uk',
  isSingleSpaced: false 
}) %}
```

This means that in almost all of the macro's use it will be a simple bit of markup like

```
{{ contactDetails() }}
```
or
```
{{ contactDetails({ isSingleSpaced: true ) }}
```

with one instance of:
```
{{
  contactDetails({
    email: 'national-operator-returns@environment-agency.gov.uk',
    isSingleSpaced: true 
  })
}}
```

PR #1107, opened 2019-12-19
Feat: Updates presentation of bill summary page

swaff:
WATER-2435

Updates the presentation of the bill summary page to align better with
the original prototype.

Includes updating the `title` macro to allow an extra large header to be
rendered.

Includes a new `heavyLine` macro which renders a 2px think hr


src/shared/assets/sass/blocks/_heavy-line.scss
```diff
@@ -0,0 +1,4 @@
+.heavy-line {
```
james-siteclick: I think this would follow our conventions if it extended and was named like the corresponding GOV.uk component, e.g.:
```
.section-break {
  @extend .gov-uk-section-break;
}
.section-break__heavy {
  @extend .govuk-section-break--visible;
  border-bottom-height: 2px;
}
```

PR #1123, opened 2020-01-08
Move error plugin into shared folder.

DGajewska:
bugfix/WATER-2455

Merge functionality into one shared plugin. Create handleUnauthorized to handle different approaches across internal/external.


server-external.js
```diff
@@ -71,6 +71,12 @@ async function start () {
       options: {
         reauthenticate: connectors.idm.users.reauthenticate.bind(connectors.idm.users)
       }
+    }, {
+      plugin: require('shared/plugins/error'),
+      options: {
+        logger,
+        contextDefaults: require('./src/external/lib/view').contextDefaults
```
james-siteclick: Is it still needed to pass these in? 
It might be possible to just use request.view in the error plugin handler now.

src/shared/plugins/error/index.js
```diff
@@ -27,28 +25,29 @@ const isUnauthorized = request => {
...
+    pluginOptions.logger.info(pick(res, ['error', 'message', 'statusCode', 'stack']));
     return request.logOut();
   }
 
-  // Unauthorised - redirect to sign in
+  // Unauthorised - redirect to welcome
   if (isUnauthorized(request)) {
-    logger.info(pick(res, ['error', 'message', 'statusCode', 'stack']));
-    return h.redirect('/signin');
+    pluginOptions.logger.info(pick(res, ['error', 'message', 'statusCode', 'stack']));
```
james-siteclick: I think it may be neater if this piece of logic moved to a handler in the auth plugin, as then there is no dependency between the plugins

PR #1125, opened 2020-01-09
Feat: Back link on batch summary from list

swaff:
WATER-2435

Missing requirement for an earlier story, this adds the back link to the
batch summary, if the user was previously on the batch list page.

Therefore not shown as part of the batch creation flow, because this
would sent the user back to the region selection page which is turn
could cause an error because two batches are not allowed to be
generated per region at a time.


swaff:
Closing to use query string instead of http header

PR #1132, opened 2020-01-15
Feat: Transaction CSV download

DGajewska:
WATER-2377

Add connector code to get invoices for a batch. Generate CSV row for each transaction returned from CM.


src/internal/modules/billing/controller.js
```diff
@@ -228,6 +229,20 @@ const postBillingBatchConfirm = async (request, h) => {
   return h.redirect('/billing/batch/list');
 };
 
+const getTransactionsCSV = async (request, h) => {
+  const { batchId } = request.params;
+
+  const { data } = await services.water.billingBatches.getInvoicesForBatch(batchId);
+  const batch = await batchService.getBatch(batchId);
```
swaff: We have a pre handler defined that can add the batch to the request where required.

If `pre: [{ method: preHandlers.loadBatch }]` is added to the route then the batch can be found on `request.defra`

PR #1133, opened 2020-01-15
Feat: Integrate backend batch APIs

swaff:
WATER-2447

Updates the UI to get batch information from the water service instead
of a stubbed response.

The water service is still implementing the mechanism to get the value
and number of invoices in a batch so for now, for the purpose of viewing
the design, the numbers are randomised.

Then updates the batch summary to use real invoice data from the water
service.


src/internal/views/nunjucks/billing/batch-list.njk
```diff
@@ -29,20 +29,20 @@
         {% for batch in batches %}
         <tr class="govuk-table__row">
           <td class="govuk-table__cell">
-            <a class="govuk-link" href="/waiting/{{batch.eventId}}">
+            <a class="govuk-link" href="/billing/batch/{{batch.id}}/summary">
```
Stephan-Hough: Hi Mark - if we change this link to point directly to the summary page and the batch is still processing it will bypass the waiting screen and will fail to get the details of the batch then also the UI will not be displayed correctly. Unless if we add an If batch status is not processing then use /billing/batch.... else /waiting/...

PR #1161, opened 2020-02-13
Update batch status from complete to ready

DGajewska:
WATER-2514

Update batch status.
Fix tariff spelling errors.

### DIFF
--- a/src/internal/modules/billing/controller.js
+++ b/src/internal/modules/billing/controller.js
@@ -233,7 +233,7 @@ const badge = {
 
 const getBatchType = (type) => {
   const batchType = type.replace(/^\w/, c => c.toUpperCase());
-  return (batchType === 'Two_part_tarrif') ? 'Two-part tarrif' : batchType;
+  return (batchType === 'Two_part_tariff') ? 'Two-part tariff' : batchType;
 };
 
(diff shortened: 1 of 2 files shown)


src/internal/modules/billing/controller.js
```diff
@@ -233,7 +233,7 @@ const badge = {
 
 const getBatchType = (type) => {
   const batchType = type.replace(/^\w/, c => c.toUpperCase());
-  return (batchType === 'Two_part_tarrif') ? 'Two-part tarrif' : batchType;
+  return (batchType === 'Two_part_tariff') ? 'Two-part tariff' : batchType;
```
james-siteclick: If this is coming from the water service, will it be all lowercase, e.g.
```
(batchType === 'two_part_tariff') ...
```

DGajewska: Yes, good point! I just noticed the spelling mistake whilst looking for batch statuses. I'll make it lowercase

DGajewska: After a further look, the previous line was capitalising the first letter in the batchType, but as we both found this confusing, I have refactored it to make it a bit clearer

PR #1168, opened 2020-02-19
Batch list view fixes

james-siteclick:
Fixes issues in the billing batch list view:

- Hides table if no batches
- Handles case where totals are not available for a batch
- Corrects spelling of 'Building' for batch label

### DIFF
--- a/src/shared/view/nunjucks/filters/charge.js
+++ b/src/shared/view/nunjucks/filters/charge.js
@@ -1,3 +1,4 @@
+const { isNull, isUndefined } = require('lodash');
 const commaNumber = require('comma-number');
 
 /**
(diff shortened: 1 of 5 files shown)


src/shared/view/nunjucks/filters/charge.js
```diff
@@ -1,3 +1,4 @@
+const { isNull, isUndefined } = require('lodash');
```
swaff: const { isNil } = require('lodash');

src/shared/view/nunjucks/filters/charge.js
```diff
@@ -8,6 +9,9 @@ const commaNumber = require('comma-number');
  * @return {String} formatted number with £ prefix
  */
 const charge = value => {
+  if (isNull(value) || isUndefined(value)) {
+    return;
+  }
```
swaff:
```
if (isNil(value)) {
   return
}
```

PR #1183, opened 2020-02-27
Feat/water 2512 invoice view

james-siteclick:
Amends the 'batch invoice' page to use data from the updated water service API.
Transactions are grouped by:
Financial year > Licence number > Charge element
Also moves some logic (e.g. batch status badge) to Nunjucks filters for easy re-use.


james-siteclick:
Please do not merge until corresponding water service API is merged

PR #1205, opened 2020-03-11
Update: make billing batch pages consistent

DGajewska:
Updates metadata and headings on the following pages:
- bill run summary page
- invoice details page
- remove invoice account page
- cancel bill run
- confirm bill run


src/internal/views/nunjucks/billing/macros/batch-metadata.njk
```diff
@@ -0,0 +1,18 @@
+{% macro batchMetadata(batch) %}
+    <div class="govuk-grid-column-full">
+      <dl class="meta govuk-!-margin-bottom-3">
+        <div class="meta__row">
+          <dt class="meta__key black-text">Date created</dt>
```
swaff: Could this be replaced with a modifier on meta__key. I don't know what the black is trying to convey, but something like `meta__key--loud`.

PR #1221, opened 2020-03-26
View 2PT Licences Ready for Billing

Stephan-Hough:
feat/water-2597
-- new route, controller and tests
-- code refactored to share code between views


src/internal/modules/billing/controllers/two-part-tariff.js
```diff
@@ -9,32 +9,48 @@ const messages = {
...
     batch,
     licences,
-    totals,
+    totals: getTotals(licencesData),
     back: `/billing/batch/list`
   });
 };
 
+const getTwoPartTariffReview = async (request, h, action) => getTwoPartTariffAction(request, h, 'review');
+const getTwoPartTariffViewReady = async (request, h) => getTwoPartTariffAction(request, h, 'ready');
```
james-siteclick: There's a handy function in lodash called partialRight to create a new function by appending arguments - this would then become:
```
const getTwoPartTariffReview = partialRight(getTwoPartTariffAction, 'review');
const getTwoPartTariffViewReady = partialRight(getTwoPartTariffAction, 'ready');
```
However your implementation above may be more clear/explicit

src/internal/modules/billing/controllers/two-part-tariff.js
```diff
@@ -9,32 +9,48 @@ const messages = {
...
     ...request.view,
     batch,
     licences,
-    totals,
+    totals: getTotals(licencesData),
     back: `/billing/batch/list`
   });
 };
 
+const getTwoPartTariffReview = async (request, h, action) => getTwoPartTariffAction(request, h, 'review');
```
james-siteclick: I think the action argument here doesn't get used so this could be:
```
const getTwoPartTariffReview = async (request, h) => getTwoPartTariffAction(request, h, 'review');
```

src/internal/modules/billing/controllers/two-part-tariff.js
```diff
@@ -9,32 +9,48 @@ const messages = {
...
+  }), initVals);
+
+  const totals = {
+    ...totalErrors,
+    ready: licences.length - totalErrors.errors,
+    total: licences.length
+  };
+  return totals;
+};
+
```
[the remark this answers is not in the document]
Stephan-Hough: Done - clever stuff! Thanks

src/internal/modules/billing/controllers/two-part-tariff.js
Stephan-Hough: very handy to know - I followed a similar pattern in the bill-run controller so will probably leave it like this for now and can refactor with the new ticket i created.

src/internal/modules/billing/controllers/two-part-tariff.js
Stephan-Hough: done

PR #1226, opened 2020-04-01
Review TPT transaction and set billable quantity

james-siteclick:
Adds the flow for a user to view TPT matching errors for a transaction, and set a billable volume.
Also:

- Makes billing captions consistent across screens
- Adds Nunjucks macros for billing to allow for re-use across billing screens
- Uses the twoPartTariffError flag to count licences in error/OK status


src/internal/modules/billing/pre-handlers.js
```diff
@@ -25,5 +27,28 @@ const redirectToWaitingIfEventNotComplete = async (request, h) => {
     : h.redirect(`/waiting/${event.event_id}`).takeover();
 };
 
+const checkBatchStatus = async (request, h, status) => {
+  const { batch } = request.pre;
+  if (batch.status !== status) {
+    return Boom.badRequest(`Batch ${batch.id} has unexpected status ${batch.status}`);
```
swaff, line 33: I wonder if this is actually a bad request. It feels like the request was valid and understood, but the server is refusing to do anything with it. This might be a 403 forbidden where the reason stays present and is therefore not confused with an authentication/authorisation error.

Difficult to know, it is a shame there is not a code for 

1) I don't understand your request
2) I understand the request but it is against the rules
3) I understand the request but you are not allowed to action it

src/internal/views/nunjucks/billing/batch-cancel-or-confirm.njk
```diff
@@ -9,11 +9,13 @@
...
 
-      <span class="govuk-caption-l">Bill run {{  batch.billRunId }}</span>
+      <span class="govuk-caption-l">{{ batch.region.displayName }} bill run {{  batch.billRunId }}</span>
       <h2 class="govuk-heading-l">{{ secondTitle }}</h2>
-      <div class="govuk-grid-row govuk-!-margin-bottom-3">
+
         {{ batchMetadata(batch) }}
-      </div>
+
+        <br />
```
swaff, line 17: I feel like our use of line breaks is a bit naughty. Would it be better to wrap the output from `{{ batchMetadata(batch) }}` in a div and apply a govuk overridden bottom margin?

src/internal/views/nunjucks/billing/two-part-tariff-quantities.njk
```diff
@@ -0,0 +1,130 @@
...
+          </td>
+          <td class="govuk-table__cell">{{ transaction.calculatedVolume ~ ' ML' if transaction.calculatedVolume | isFinite }}</td>
+          <td class="govuk-table__cell">{{ error }}</td>
+        </tr>
+      </tbody>
+    </table>
+
+ 
+
+    <br />
```
swaff, line 115: Same here

PR #1301, opened 2020-06-05
Licence rename connector method for CRM moved to water service API

Stephan-Hough:
Feat/water 2644 rename licence
This change is to record event data when a licence is re/named and to shift the CRM call to the water service


src/external/modules/view-licences/controller.js
```diff
@@ -100,11 +101,12 @@ async function postLicenceRename (request, h) {
 
   // Rename licence
   const { name } = getValues(form);
-  const { error } = await services.crm.documents.setLicenceName(documentId, name);
+  const rename = !!documentName;
```
james-siteclick, line 104: The rename param would not be needed here if this is changed in the water service

src/external/modules/view-licences/controller.js
```diff
@@ -100,11 +101,12 @@ async function postLicenceRename (request, h) {
...
+  const { error } = await services.water.documents.postLicenceRename(documentId, { documentName: name, rename, userName });
```
james-siteclick, line 105: I'm not sure if the { error } = ...  and throwIfError are doing anything here.
For example, if the water service returns a 404 error, I think an error would already be thrown by the services.water.documents.postLicenceRename() call.  So it may be OK to just have:
```
await services.water.documents.postLicenceRename(documentId, { documentName: name, rename, userName });
```
Or if you want to catch it:
```
try {
  await services.water.documents.postLicenceRename(documentId, { documentName: name, rename, userName });
}
catch(err) {
  // do something else
}

src/external/modules/view-licences/controller.js
Stephan-Hough: change applied

src/external/modules/view-licences/controller.js
Stephan-Hough: done

PR #1354, opened 2020-08-03
Add a contact for invoice accounts flow

Stephan-Hough:
feat/water-2722
-- prehandler for company added
-- routes and controllers refactored into sperate files
-- new add contact routes and controller
-- bug fix on route validation failure when returning empty form values
-- new select and create contact forms
-- tests added for new routes, forms and controller


timtamimi (changes requested):
Code seems sensible! Just have a two UI/UX comments:

1. On the 'Select an existing contact...' page,

Screenshot 2020-08-25 at 12 50 09

I don't think the prompt 'Select' makes sense here. In the screenshot above, there are no existing contacts for this particular license/company.

_I appreciate this is a UXD issue rather than a development matter._


2. Less important, but potentially still worth visiting, is on the 'Do you need to add an FAO' prompt as seen below:

Screenshot 2020-08-25 at 12 49 52

In the screenshot above, The options listed are 'No' and 'Yes'. The Prototype has them the other way around, 'Yes' followed by 'No'.

timtamimi (approved):
Looks good to me. Thanks for making the tweaks. 👍🏽

PR #1357, opened 2020-08-06
Feat: Flow to enter a new address

DGajewska:
WATER-2719


src/internal/modules/address-entry/controller.js
```diff
@@ -0,0 +1,98 @@
...
+    back: helpers.getPostcodeUrl(request),
+    postcode: helpers.getPostcode(request),
+    form: sessionForms.get(request, selectAddress.form(request))
+  });
+};
+
+const postSelectAddress = (request, h) => {
+  const { uprn } = request.payload;
+  const form = forms.handleRequest(
+    selectAddress.form(request, uprn),
```
james-siteclick, line 56: This may be a silly question, but do all the addresses in this DB have a UPRN?

src/internal/modules/address-entry/controller.js
```diff
@@ -0,0 +1,98 @@
...
+  let form = forms.handleRequest(
+    manualAddressEntry.form(request, request.payload),
+    request,
+    manualAddressEntry.schema
+  );
+
+  form = manualAddressEntry.applyRequiredFieldErrors(form, request.payload);
+
+  if (form.isValid) {
+    return storeAddressAndRedirect(request, h, omit(request.payload, ['csrfToken']));
```
james-siteclick, line 88: I assume this would overwrite any uprn already in the session from a previous request to the other form?  
This would be desirable because we wouldn't want them to select an address, then go and edit a custom address so the address didn't match the UPRN.

src/internal/modules/address-entry/controller.js
```diff
@@ -0,0 +1,98 @@
...
+
+  if (form.isValid) {
+    return h.redirect(helpers.getSelectAddressUrl(request));
+  }
+  const { redirectPath, back } = request.yar.get(helpers.SESSION_KEY);
+  return h.postRedirectGet(form, '/address-entry/postcode', { redirectPath, back });
+};
+
+const getSelectAddress = (request, h) => {
+  helpers.setAddressSearchResults(request);
```
james-siteclick, line 42: I think I'd be inclined to put the postcode in a hidden field rather than store the whole address search result set in the session.
This makes the POST handler stateless which is always desirable.
The POST handler would need to make another request to the water service with the postcode and then find the address by UPRN.
To save API requests to the address service, which cost, Redis backed cache could be implemented in the water service, and the UI would not need to worry.

src/internal/modules/address-entry/forms/select-address.js
```diff
@@ -0,0 +1,70 @@
...
+    label: 'Select an address',
+    choices: getAddressChoices(addressSearchResults)
+  }, parseInt(uprn)));
+
+  f.fields.push(fields.link(null, {
+    text: 'I cannot find the address in the list',
+    url: '/address-entry/manual-entry?country=United Kingdom'
+  }));
+
+  f.fields.push(fields.hidden('csrf_token', {}, csrfToken));
```
james-siteclick, line 58: I think it would be good to add a hidden field for the postcode rather than putting in the session data.
This can be obtained from `request.query.postcode`

src/internal/modules/address-entry/forms/uk-postcode.js
```diff
@@ -0,0 +1,47 @@
...
+ * Creates an object to represent the form for capturing the
+ * UK postcode to use for looking up addresses
+ *
+ * @param {Object} request The Hapi request object
+ * @param {String} postcode The UK postcode
+ */
+const form = (request, postcode) => {
+  const { csrfToken } = request.view;
+
+  const f = formFactory('/address-entry/postcode');
```
james-siteclick, line 16: This form could have a GET method (would not need CSRF field in this case)

src/internal/modules/address-entry/routes.js
```diff
@@ -0,0 +1,129 @@
...
+        }
+      }
+    },
+
+    getSelectAddress: {
+      method: 'GET',
+      path: '/address-entry/address/select',
+      handler: controller.getSelectAddress,
+      options: {
+        auth: {
```
james-siteclick, line 60: This route can be modified to accept:
```
query : {
  postcode : Joi.string().required().allow('')
}
```

src/internal/modules/address-entry/routes.js
```diff
@@ -0,0 +1,129 @@
...
+            licenceId: Joi.string().guid().optional().allow(null),
+            back: Joi.string().required(),
+            redirectPath: Joi.string().required(),
+            form: Joi.string().optional()
+          }
+        }
+      }
+    },
+
+    postPostcode: {
```
james-siteclick, line 38: I don't think this route is needed

james-siteclick (approved):

PR #1397, opened 2020-09-17
Feat: Non chargeable licence flow

swaff:
WATER-2689

Updates the UI to enable the non chargeable licence flow from the charge
information tab.

Makes the start-date form dual purpose, now capturing either the start
date of a new charge version, or the effective date for a non chargeable
charge version.

Depends on https://github.com/DEFRA/water-abstraction-service/pull/1144

### DIFF
--- a/src/internal/modules/charge-information/controller.js
+++ b/src/internal/modules/charge-information/controller.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { get } = require('lodash');
+
 const forms = require('./forms');
 const actions = require('./lib/actions');
 const routing = require('./lib/routing');
@@ -8,7 +10,7 @@ const services = require('../../lib/connectors/services');
 
 const getLicencePageUrl = async licence => {
   const document = await services.crm.documents.getWaterLicence(licence.licenceNumber);
-  return `/licences/${document.document_id}`;
+  return `/licences/${document.document_id}#charge`;
 };
 
 /**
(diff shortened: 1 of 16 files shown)


src/internal/modules/charge-information/controller.js
```diff
@@ -105,47 +107,106 @@ const postUseAbstractionData = createPostHandler(
   request => routing.getCheckData(request.pre.licence)
 );
 
-const getCheckData = async (request, h) => {
-  const { licence, draftChargeInformation } = request.pre;
+const getInvoiceAccountAddressFromDraft = draftChargeInformation => {
```
james-siteclick, line 110: I would expect we would have something in the session along the lines of:
```
{
  "chargeVersion" : {
    "invoiceAccount" : {
        "id" : "43443-434-43-334-3443-4"
     },
     ...
  }
```
This single invoice account ID should be enough to get the invoice account from the crm via the water service including its currently selected address.

I'm assuming the data in `billingAccount.billingAccount...` has been collected during the "select invoice account" sub-flow, but if a charge version is in review and loaded back from the workflow table by a different user, none of this would be populated.

src/internal/modules/charge-information/controller.js
```diff
@@ -105,47 +107,106 @@ const postUseAbstractionData = createPostHandler(
...
 };
 
 const postCheckData = async (request, h) => {
   const { nextStep } = request.payload;
-  const { licence } = request.pre;
+  const { licence, isChargeable } = request.pre;
 
   if (nextStep === 'confirm') {
     // send the draft data to the water service to save
     console.log('Collected data', request.pre.draftChargeInformation);
```
james-siteclick, line 143: It looks from this as though the data is being collected in a Redis key not using request.yar.

I had started building this a while ago before we decided to use a PG table in the water service for persistence of the workflow data, and this provided a way for it to be shared amongst internal users.

So there's no need to do this anymore I don't think.  All data in the flow can be stored in the user's session as normal with request.yar.

james-siteclick (changes requested):

Stephan-Hough:
new PR created https://github.com/DEFRA/water-abstraction-ui/pull/1417 closing this one

PR #1394, opened 2020-09-16
Feat/water 2837 charge information 4.1 select existing company

timtamimi:
Original ticket here: https://eaflood.atlassian.net/browse/WATER-2837

PR #1409, opened 2020-09-28
Contact entry module - refactored

timtamimi:
This is a refactored solution for this ticket: https://eaflood.atlassian.net/browse/WATER-2837

The original attempt/solution is here: https://github.com/DEFRA/water-abstraction-ui/pull/1394

To be merged in conjunction with https://github.com/DEFRA/water-abstraction-tactical-crm/pull/633

### DIFF
--- a/src/internal/modules/contact-entry/routes.js
+++ b/src/internal/modules/contact-entry/routes.js
@@ -54,6 +54,20 @@ module.exports = {
       }
     }
   },
+  getAddressEntered: {
+    // Route for handling the handover from the address entry module
+    method: 'GET',
+    path: '/contact-entry/new/address-entered',
+    handler: controllers.getAddressEntered,
+    options: {
+      validate: {
+        query: {
+          sessionKey: Joi.string().uuid().required(),
+          form: Joi.string().optional()
+        }
+      }
+    }
+  },
   postCompanySearch: {
     // Route for posting the name or number of a new company contact
     // It searches companies house for a matching company
--- a/src/internal/modules/contact-entry/routes.js
+++ b/src/internal/modules/contact-entry/routes.js
@@ -1,0 +1,116 @@
+const preHandlers = require('./pre-handlers');
+const Joi = require('@hapi/joi');
+const controllers = require('./controllers');
+
+module.exports = {
+  getNew: {
+    // Route for starting the flow. Registers data in session.
+    method: 'GET',
+    path: '/contact-entry/new',
+    handler: controllers.getNew,
+    options: {
+      validate: {
+        query: {
+          sessionKey: Joi.string().uuid().required(),
+          searchQuery: Joi.string().optional(),
+          back: Joi.string().optional(),
+          redirectPath: Joi.string().required(),
+          form: Joi.string().optional()
+        }
+      }
+    }
+  },
+  getSelectAccountType: {
+    // Route for displaying the page for selecting an account type when creating a new contact
+    method: 'GET',
+    path: '/contact-entry/new/account-type',
+    handler: controllers.getSelectAccountTypeController,
+    options: {
+      validate: {
+        query: {
+          sessionKey: Joi.string().uuid().required(),
+          form: Joi.string().optional()
+        }
+      }
+    }
+  },
+  postSelectAccountType: {
+    // Route for selecting an account type when creating a new contact
+    method: 'POST',
+    path: '/contact-entry/new/account-type',
+    handler: controllers.postSelectAccountTypeController
+  },
+  getEnterNewDetails: {
+    // Route for displaying the list of existing contact addresses.
+    method: 'GET',
+    path: '/contact-entry/new/details',
+    handler: controllers.getDetailsController,
+    options: {
+      validate: {
+        query: {
+          sessionKey: Joi.string().uuid().required(),
+          form: Joi.string().optional()
+        }
+      }
+    }
+  },
+  postCompanySearch: {
+    // Route for posting the name or number of a new company contact
+    // It searches companies house for a matching company
+    method: 'POST',
+    path: '/contact-entry/new/details/company-search',
+    handler: controllers.postCompanySearchController
+  },
+  getSelectCompany: {
+    method: 'GET',
+    path: '/contact-entry/new/details/company-search/select-company',
+    handler: controllers.getSelectCompanyController,
+    options: {
+      pre: [
+        { method: preHandlers.searchForCompaniesInCompaniesHouse, assign: 'companiesHouseResults' }
+      ],
+      validate: {
+        query: {
+          sessionKey: Joi.string().uuid().required(),
+          form: Joi.string().optional()
+        }
+      }
+    }
+  },
+  postSelectCompany: {
+    method: 'POST',
+    path: '/contact-entry/new/details/company-search/select-company',
+    handler: controllers.postSelectCompanyController,
+    options: {
+      pre: [
+        { method: preHandlers.searchForCompaniesInCompaniesHouse, assign: 'companiesHouseResults' }
+      ]
+    }
+  },
+  getSelectCompanyAddress: {
+    method: 'GET',
+    path: '/contact-entry/new/details/company-search/select-company-address',
+    handler: controllers.getSelectCompanyAddressController,
+    options: {
+      pre: [
+        { method: preHandlers.returnCompanyAddressesFromCompaniesHouse, assign: 'companiesHouseAddresses' }
+      ],
+      validate: {
+        query: {
+          sessionKey: Joi.string().uuid().required(),
+          form: Joi.string().optional()
+        }
+      }
+    }
+  },
+  postSelectCompanyAddress: {
+    method: 'POST',
+    path: '/contact-entry/new/details/company-search/select-company-address',
+    handler: controllers.postSelectCompanyAddressController,
+    options: {
+      pre: [
+        { method: preHandlers.returnCompanyAddressesFromCompaniesHouse, assign: 'companiesHouseAddresses' }
+      ]
+    }
+  }
+};
--- a/src/internal/modules/invoice-accounts/controllers/contacts.js
+++ b/src/internal/modules/invoice-accounts/controllers/contacts.js
@@ -4,9 +4,12 @@ const dataService = require('../services/data-service');
 const helpers = require('../lib/helpers');
 const { selectContactForm, selectContactFormSchema } = require('../forms/select-contact');
 const { createContactForm, createContactFormSchema } = require('../forms/create-contact');
+const { selectExistingContactForm, selectExistingContactSchema } = require('../forms/select-existing-contact');
 const sessionForms = require('shared/lib/session-forms');
 const forms = require('shared/lib/forms');
+const queryString = require('querystring');
 const urlJoin = require('url-join');
+const uuid = require('uuid');
 const { has, isEmpty, omit, pickBy } = require('lodash');
 
 /**
@@ -55,7 +58,7 @@ const getContactCreate = async (request, h) => {
   const { regionId, companyId } = request.params;
   const session = dataService.sessionManager(request, regionId, companyId);
   const selectedContact = isEmpty(session.contact) ||
-  (has(session.contact, 'department') && !has(session.contact, 'firstName')) ? {} : session.contact;
+    (has(session.contact, 'department') && !has(session.contact, 'firstName')) ? {} : session.contact;
 
   return h.view('nunjucks/form', {
     ...request.view,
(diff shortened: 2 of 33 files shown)


src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
```diff
@@ -22,27 +23,28 @@ const helpers = require('../lib/helpers');
...
   const { licenceId, redirectPath } = request.query;
-  const { company } = request.pre;
+  const { company, companies } = request.pre;
   // licenceId is an optional query param. if supplied it will be displayed as the caption on the forms for the session
   const { licenceNumber } = licenceId ? await dataService.getLicenceById(licenceId) : { licenceNumber: null };
   // The company name and licence number set here will be used in the select address page
   const data = { viewData: { redirectPath, licenceNumber, licenceId, companyName: titleCase(company.name) } };
   const session = dataService.sessionManager(request, regionId, companyId, data);
-  const selectedCompany = has(session, 'agent') && isEmpty(session.agent) ? company : helpers.getAgentCompany(session);
+  const selectedCompany = has(session, 'agent') && isEmpty(session.agent) ? company : await helpers.getAgentCompany(session);
```
Stephan-Hough, line 32: The loadCompanies pre handler already fetches the agent company if it exists in the session. I would rather leave it here than in the pre-handler because having an agent is more of an exception than not so the prehandler will in most cases do less work

src/internal/modules/invoice-accounts/pre-handlers.js
```diff
@@ -2,9 +2,32 @@
 
 const Boom = require('boom');
 const { partial } = require('lodash');
-
+const dataService = require('./services/data-service');
+const { has } = require('lodash');
+const helpers = require('./lib/helpers');
 const { water } = require('../../lib/connectors/services');
 
+const loadCompanies = async (request) => {
```
Stephan-Hough, line 10: The loadCompanies, loadBillingContact and getCompany methods all get the similar data. If you stick with the one getCompany and only get the agent company when there is an agent it will be less code and avoid duplication.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
```diff
@@ -22,27 +23,28 @@ const helpers = require('../lib/helpers');
...
     back: '/manage',
-    form: sessionForms.get(request, selectCompanyForm(request, company, selectedCompany))
+    form: sessionForms.get(request, selectCompanyForm(request, companies, selectedCompany))
   });
 };
 
 const postCompany = async (request, h) => {
   const { regionId, companyId } = request.params;
-  const { company } = request.pre;
+  const { companies } = request.pre;
```
Stephan-Hough, line 45: if you decide to get rid of the loadCompanies pre-handler then you can do the same here and only add the agent company if it has been selected. i.e. if the payload.selected company id != company Id then get it using the helpers.getAgentCompany.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
```diff
@@ -53,35 +55,74 @@ const postCompany = async (request, h) => {
 
 const getAddress = async (request, h) => {
   const { regionId, companyId } = request.params;
-  const addresses = await dataService.getCompanyAddresses(companyId);
   // get the session data to check if the address has been set and if it is new or existing
   const session = dataService.sessionManager(request, regionId, companyId);
+  const addresses = await helpers.getAllAddresses(companyId, session);
```
Stephan-Hough, line 60: this helper method gets all the addresses for the company and the agent company. If the user selected the agent company they could potentially select an address from the original company and I am not sure we would want this to happen. In the database the address could then be linked to the original company and the agent company. 

I think it would be better to only load the addresses for the selected company into this form

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
```diff
@@ -53,35 +55,74 @@ const postCompany = async (request, h) => {
...
   // @TODO this might need a mapper to map the session address data to the Company address shape passed to the form
```
Stephan-Hough, line 61: can we remove this comment now?

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
```diff
@@ -53,35 +55,74 @@ const postCompany = async (request, h) => {
...
   const { regionId, companyId } = request.params;
-  const addresses = await dataService.getCompanyAddresses(companyId);
+  const session = dataService.sessionManager(request, regionId, companyId);
+  const addresses = await helpers.getAllAddresses(companyId, session);
   const schema = selectAddressFormSchema(request.payload);
   const form = forms.handleRequest(selectAddressForm(request, addresses), request, schema);
   if (form.isValid) {
     const { selectedAddress } = forms.getValues(form);
-    dataService.sessionManager(request, regionId, companyId, { address: { addressId: selectedAddress } });
+    if (selectedAddress !== tempId) {
```
Stephan-Hough, line 81: what if the selected address does have a tempId?

src/internal/modules/invoice-accounts/lib/helpers.js
```diff
@@ -11,11 +13,15 @@ const getFormTitleCaption = (licenceNumber) => {
 const processCompanyFormData = (request, regionId, companyId, formData) => {
   const { selectedCompany, companySearch } = forms.getValues(formData);
   if (selectedCompany === 'company_search') {
-    // TODO place holder -- route does not exist
-    return `company-search?filter=${companySearch}`;
+    return `contact-search?filter=${companySearch}`;
   } else {
     const agentId = selectedCompany === companyId ? null : selectedCompany;
-    dataService.sessionManager(request, regionId, companyId, { agent: agentId });
+    if (agentId) {
```
Stephan-Hough, line 19: I think the above line was fine as it is. You only need to set the agent either to null or if the selected company id is not the same as the main company id then set the agent = to the selected company id.

src/internal/modules/invoice-accounts/lib/helpers.js
```diff
@@ -11,11 +13,15 @@ const getFormTitleCaption = (licenceNumber) => {
 const processCompanyFormData = (request, regionId, companyId, formData) => {
   const { selectedCompany, companySearch } = forms.getValues(formData);
   if (selectedCompany === 'company_search') {
-    // TODO place holder -- route does not exist
-    return `company-search?filter=${companySearch}`;
+    return `contact-search?filter=${companySearch}`;
```
Stephan-Hough, line 16: This could potentially lead to confusion if we call it contact search because you are basically searching for a company and not a contact in the database.

src/internal/modules/invoice-accounts/lib/helpers.js
```diff
@@ -54,18 +60,47 @@ const getSelectedAddress = async (companyId, session) => {
...
+    const addresses = await getAllAddresses(companyId, session);
     const selectedAddress = addresses.find(address => (address.id === session.address.addressId));
     return selectedAddress;
   };
 };
 
-const getAgentCompany = (session) => {
-  if (has(session, 'agent')) {
-    return session.agent.companyId === tempId ? session.agent : dataService.getCompany(session.agent.companyId);
+const getAllAddresses = async (companyId, session) => {
```
Stephan-Hough, line 69: I dont think you need this method based on my previous comments

src/internal/modules/invoice-accounts/lib/helpers.js
```diff
@@ -54,18 +60,47 @@ const getSelectedAddress = async (companyId, session) => {
...
 };
 
+const getCompanyName = async (request) => {
+  const { sessionKey } = request.query;
+  const { companyId } = request.params;
+  let currentState = await request.yar.get(sessionKey);
+  if (currentState.newCompany) {
+    return currentState.agent.name;
+  } else {
+    const originalCompany = await dataService.getCompany(companyId);
```
Stephan-Hough, line 99: this company name is already in the session viewData object

src/internal/modules/invoice-accounts/controllers/contacts.js
```diff
@@ -84,8 +87,49 @@ const postContactCreate = async (request, h) => {
   return h.postRedirectGet(form, urlJoin('/invoice-accounts/create/', regionId, companyId, 'create-contact'));
 };
 
+const getContactSearch = async (request, h) => {
```
Stephan-Hough, line 90: I think this should be called getSearchCompany and move to the invoice-accounts controller because you are searching for a company in the database and not a contact.

src/internal/modules/invoice-accounts/controllers/contacts.js
```diff
@@ -84,8 +87,49 @@ const postContactCreate = async (request, h) => {
...
+  });
+};
+
+const postContactSearch = async (request, h) => {
+  const { id, filter } = request.payload;
+  const { regionId, companyId } = request.params;
+  const form = forms.handleRequest(selectExistingContactForm(request), request, selectExistingContactSchema);
+  if (form.isValid) {
+    if (id === 'new') {
+      let volatileKey = uuid();
```
Stephan-Hough, line 110: you could change to const because the key doesn't change

src/internal/modules/invoice-accounts/lib/helpers.js
timtamimi: Not necessarily - the company name in `session.viewData` may be over-written with the name of the agent company as part of the contact entry handover. :)

src/internal/modules/invoice-accounts/controllers/contacts.js
timtamimi: Agreed. I will change this

src/internal/modules/invoice-accounts/controllers/contacts.js
timtamimi: OK. I will move this.

src/internal/modules/invoice-accounts/lib/helpers.js
timtamimi: The reason I changed the URL path is that this is visible to the end user - the language in the prototype says 'Contact' rather than 'Company'. Are you sure it should be `company-search`? Not a big deal I suppose - Happy to change it if you feel I should.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
timtamimi: I don't know - I didn't put it there 😂 

I'll remove it now anyway.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
timtamimi: Good point.

I ran this past Chris and Ruth.

The outcome was: It's OK to show both companies' addresses because it avoids duplications in data entry to the database and expedites the user journey. 

We discussed that we can change this later down the line.

(So I will leave it as it is)

src/internal/modules/invoice-accounts/pre-handlers.js
timtamimi: Fair point. I have removed the unnecessary prehandlers.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
timtamimi: I've kept the pre-handler but I tidied this up, by getting the agent using the helper above and passing the prehandler once.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
timtamimi: I have slimmed this down by removing one of the pre-handlers.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
timtamimi: `selectedAddress` here is not an address object - rather it's either a GUID or a string that says `new_address`. It will never be null because Joi shouldn't allow it. I hope that makes sense.

I will add a comment to explain this in the code. :)

src/internal/modules/invoice-accounts/lib/helpers.js
timtamimi: > I think the above line was fine as it is.

The purpose of the change is to: 
1. inject the name of the agent company into the viewData object, so that the agent company name appears in the UI under `Select an address for COMPANY NAME`
2. prevent the company from being entered as its own agent

> on line 21 below we don't change the companyId because this is the original and main company for the licence.

Ooo. I think I disagree: I think the agent object should only represent the agent (if there is one). The original company ID can always be derived from the request parameters, so there's no need to put the original companyId in the session.agent object.

src/internal/modules/contact-entry/routes.js
```diff
@@ -0,0 +1,130 @@
+const preHandlers = require('./pre-handlers');
+const Joi = require('@hapi/joi');
+const controllers = require('./controllers');
+
+module.exports = {
+  getNew: {
```
Stephan-Hough, line 6: I think this is an unnecessary route and can be merged with getSelectAccountType.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
```diff
@@ -22,66 +25,143 @@ const helpers = require('../lib/helpers');
...
+      const path = `/contact-entry/new?` + queryString.stringify({
+        searchQuery: filter,
+        sessionKey: volatileKey,
+        back: `/invoice-accounts/create/${regionId}/${companyId}/contact-search`,
+        redirectPath: `/invoice-accounts/create/${regionId}/${companyId}/contact-entry-complete?sessionKey=${volatileKey}`
+      });
+      return h.redirect(path);
+    } else {
+      // Set the display company name
+      await dataService.sessionManager(request, regionId, companyId, { agent: { companyId: id } });
```
Stephan-Hough, line 88: You need to add the agent company name here to the view data

src/internal/modules/contact-entry/routes.js
timtamimi: I agree completely that this is a useless route at the moment. 

I added it for future-proofing, so that in the future when we want to call the contact-entry module, we can pre-load it with a config object that states whether the selected object should be:
1. New, existing or either
2. A person, a company, or either

This would pre-populate the session object and then skip into the right parts of the flow. I hope this makes sense.

src/internal/modules/invoice-accounts/controllers/invoice-accounts.js
timtamimi: You're totally right. My apologies.

Stephan-Hough (approved):

PR #1416, opened 2020-10-06
Feat/water 2777 end agreements

timtamimi:
https://eaflood.atlassian.net/browse/WATER-2777


src/internal/modules/agreements/controller.js
```diff
@@ -30,5 +35,73 @@ const postDeleteAgreement = async (request, h) => {
...
+};
+
+const getConfirmEndAgreement = async (request, h) => {
+  const { agreement, licence, document } = request.pre;
+  const { agreementId } = request.params;
+  const { endDate } = await helpers.sessionManager(request, agreementId);
+  return h.view('nunjucks/agreements/end', {
+    ...getDefaultView(request),
+    pageTitle: 'You\'re about to end this agreement',
+    back: `/licences/${document.document_id}#charge`,
```
DGajewska, line 81: I think this should be going back to the previous page, where the user enters the end date, as opposed to the charge info tab

test/internal/modules/agreements/controller.js
```diff
@@ -22,11 +22,18 @@ const createRequest = () => ({
   params: {
     agreementId: 'test-agreement-id'
   },
+  yar: {
+    get: sandbox.stub(),
+    set: sandbox.stub()
+  },
```
DGajewska, line 28: I don't think you need to stub yar if you're stubbing the `sessionManager` helpers method

DGajewska (changes requested):

src/internal/views/nunjucks/agreements/end.njk
```diff
@@ -0,0 +1,60 @@
...
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+
+{% block content %}
+
+  {{ title(pageTitle, caption) }}
+  {{ govukWarningText({
+    text: 'When you end this agreement, the licence will be added 
+    to the supplementary bill run. A new charge will be calculated 
+    for the time the agreement was in place.'
+  })}}
```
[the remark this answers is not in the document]
timtamimi: You might be thinking of validation error messages which are normally in the two-thirds divs? Whereas this is meant to span the full width.

Screenshot 2020-10-07 at 09 24 58

src/internal/modules/agreements/controller.js
timtamimi: Fair point. I'll change it

test/internal/modules/agreements/controller.js
timtamimi: Yeah - `yar.get` and `yar.set` need to be stubbed because `session-forms` uses those methods and isn't in itself stubbed :D

DGajewska: Ah yes, you're right

DGajewska (approved):

PR #1434, opened 2020-10-16
User management feature

timtamimi:
- Adds a page that lists internal users.
- Adds a link to the page mentioned above into the manage tab
- Adds a reinstate user controller
- Adds reinstate user forms/pages
- Adds relevant tests
- Modifies 'Delete' to 'Disable' in the UI, to reflect the purpose of the feature a bit more accurately.

Paired with https://github.com/DEFRA/water-abstraction-service/pull/1187

### DIFF
--- a/src/internal/modules/account/controller.js
+++ b/src/internal/modules/account/controller.js
@@ -1,6 +1,7 @@
 const { createUserForm, createUserSchema } = require('./forms/create-user');
 const { setPermissionsForm, setPermissionsSchema } = require('./forms/set-permissions');
 const { deleteUserForm, deleteUserSchema } = require('./forms/delete-user');
+const { reinstateUserForm, reinstateUserSchema } = require('./forms/reinstate-user');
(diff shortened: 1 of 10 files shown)


src/internal/modules/account/controller.js
```diff
@@ -95,7 +101,7 @@ const getDeleteUserAccount = async (request, h, formFromPost) => {
     ...request.view,
     userEmail,
     form,
-    back: `/user/${userId}/status`
+    back: `/accounts`
```
DGajewska, line 104: You can get to this page from 2 different places, this might need to be more flexible to reflect this

DGajewska (changes requested):
I think that there's some more work around updating the language i.e. there is still a link on the user page to 'delete user account'.
Also, the internal users page might need pagination.

timtamimi:
Spoken to UCD about this and a decision was made against this change, so I'm closing this PR.

PR #1438, opened 2020-10-19
UI Corrections following QA feedback

timtamimi:
In response to: https://eaflood.atlassian.net/browse/WATER-2837?focusedCommentId=247488

### DIFF
--- a/src/internal/modules/contact-entry/forms/company-search-select-company.js
+++ b/src/internal/modules/contact-entry/forms/company-search-select-company.js
@@ -7,12 +7,13 @@ const getCompanyText = row => {
 };
 
 const getCompanyChoices = companies => {
-  const choices = companies.map(record => {
-    return ({
-      value: record.company.companyNumber,
-      label: getCompanyText(record)
-    });
-  });
+  const choices = [{ value: null, label: `${companies.length} companies found` },
+    ...companies.map(record => {
+      return ({
+        value: record.company.companyNumber,
+        label: getCompanyText(record)
+      });
+    })];
   return choices;
 };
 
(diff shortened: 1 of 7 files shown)


src/internal/modules/contact-entry/forms/company-search-select-company.js
```diff
@@ -7,12 +7,13 @@ const getCompanyText = row => {
 };
 
 const getCompanyChoices = companies => {
-  const choices = companies.map(record => {
-    return ({
-      value: record.company.companyNumber,
-      label: getCompanyText(record)
-    });
-  });
+  const choices = [{ value: null, label: `${companies.length} companies found` },
```
james-siteclick, line 10: Does the first choice with `null` value here get rendered with a radio button?

src/internal/modules/contact-entry/forms/company-search-select-company.js
```diff
@@ -7,12 +7,13 @@ const getCompanyText = row => {
...
+    ...companies.map(record => {
+      return ({
+        value: record.company.companyNumber,
```
james-siteclick, line 13: I assume the data for this is comping from companies house?  (just thinking that the CRM company `companyNumber` property is null for NALD data)

src/internal/modules/contact-entry/forms/company-search-select-company.js
timtamimi: Good question.
It's a dropdown menu rather than radio, and the null is intended to ensure that the validation kicks in, so that you can't accidentally select that first <option> which is intended to display the result count.

Screenshot 2020-10-19 at 13 53 11

src/internal/modules/contact-entry/forms/company-search-select-company.js
timtamimi: This is correct. This array of options comes from Companies House API.

james-siteclick (approved):

PR #1442, opened 2020-10-20
Charge version workflow API / scope bug

james-siteclick:
Fixes issue where envelope of charge version workflow API response not parsed correctly


timtamimi (approved):
Tested on preprod by pulling the branch manually. Looks good.

PR #1446, opened 2020-10-20
WATER-2853 paper forms UI

james-siteclick:
Partially implements `water-2853` UI.

Provides pages for:

* Initial search by licence number(s)
* Check your answers page
* Select address page
* Select returns page

Implementation is via a reducer pattern which manages the session state during the flow.


src/internal/modules/returns-notifications/controllers/paper-forms.js
```diff
@@ -0,0 +1,127 @@
...
+};
+
+const isLicenceHolderRole = role => role.roleName === crmRoles.licenceHolder;
+
+const mapReturnToView = ret => ({
+  legacyId: ret.returnRequirement.legacyId,
+  details: getReturnStatusString(ret)
+});
+
+const mapStateToView = state => Object.values(state).map(({ document, returns, licence, selectedRole }) => ({
```
Stephan-Hough, line 83: This could be moved to a helpers or mapper file in the lib folder as this controller might grow more?

src/internal/views/nunjucks/returns-notifications/check-answers.njk
```diff
@@ -0,0 +1,83 @@
...
+              </dd>
+            </div>
+          </dl>
+
+        {% endfor %}
+
+        {{ formRender(form)}}
+      </div>
+    </div>
+  {% endblock %}
```
Stephan-Hough, line 83: Need to add a new line here.

src/shared/lib/mappers/company.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+const mapCompanyToString = company =>
```
Stephan-Hough, line 3: Is this required when you only getting the name from the company?

Stephan-Hough (changes requested):
Added a couple of comments on the files but some minor observation is I would prefer to keep the main routes file in the root directory for the module to remain consistent with the other modules and the same with the pre-handler, not sure this should be grouped with the controllers.

src/shared/lib/mappers/company.js
james-siteclick: Even though this logic is very simple at the moment, I think it is best to have the logic in one central location where it can easily be modified rather than repeated around the project

Stephan-Hough (approved):

PR #1455, opened 2020-10-23
Feat/water 2885 charge info workflow 3

timtamimi:
Ticket: https://eaflood.atlassian.net/browse/WATER-2885

Linked to: https://github.com/DEFRA/water-abstraction-service/pull/1201


src/internal/modules/charge-information/controllers/view-charge-information.js
```diff
@@ -28,20 +33,71 @@ const getViewChargeInformation = async (request, h) => {
...
+    } else {
+      await services.water.chargeVersionWorkflows.patchChargeVersionWorkflow(
+        request.params.chargeVersionWorkflowId,
+        request.payload.reviewOutcome,
+        request.payload.reviewerComments,
+        {}
+      );
+    }
+    const { document_id: documentId } = await services.water.licences.getDocumentByLicenceId(licence.id);
+    return h.redirect(`/licences/${documentId}#charge`);
```
Stephan-Hough, line 97: you need to clear the session cache when you exit this flow.

src/internal/modules/charge-information/forms/review.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const reviewFormSchema = () => ({
+  csrf_token: Joi.string().uuid().required(),
+  approval_notice: Joi.any(),
+  reviewOutcome: Joi.string().required().allow('approve', 'changes_requested'),
+  reviewerComments: Joi.string().when('reviewOutcome', {
+    is: 'changes_requested',
+    then: Joi.string().required(),
+    otherwise: Joi.string().optional().allow('').allow(null).default('')
```
Stephan-Hough, line 58: I don't think you need this line. See the start date form for charge info. 
src/internal/modules/charge-information/forms/start-date.js

src/internal/modules/charge-information/lib/routing.js
```diff
@@ -23,6 +23,8 @@ exports.getCreateBillingAccount = (licence, licenceHolderRole, redirect) => {
   return `/invoice-accounts/create/${region.id}/${licenceHolderRole.company.id}?${qs}`;
 };
 
+exports.postReview = request => `/licences/${request.params.licenceId}/charge-information/${request.params.chargeVersionWorkflowId}/review`;
```
Stephan-Hough, line 26: It might be better to remain consistent with the other routes and use the creatUrl() method:

exports.postReview = (chargeVersionWorkflowId, licenceId) => createUrl(`${chargeVersionWorkflowId}/review`)(licenceId);

You will also need to change the action in the form to get the licenceId and cvWorkflowId form the request object to pass it in here

src/internal/modules/charge-information/forms/review.js
timtamimi: I'm afraid this line is necessary, because the reviewerComments field is still there, even if the workflow is `approved`. Leaving it out will result in the error seen below

Screenshot 2020-10-28 at 16 51 23

src/internal/modules/charge-information/lib/routing.js
timtamimi: Cool. Done ✅

src/internal/modules/charge-information/forms/review.js
Stephan-Hough: This will work

` reviewerComments: Joi.when('reviewOutcome', {
    is: 'changes_requested',
    then: Joi.string().required()`

Stephan-Hough (approved):

PR #1458, opened 2020-10-28
Feat/water 2853 paper forms ui 4

james-siteclick:
WATER-2853 
Adds the remaining pages in the paper forms flow, including posting the data to the water service.


src/internal/modules/returns-notifications/lib/controller.js
```diff
@@ -21,15 +22,23 @@ const checkAnswersRoute = '/returns-notifications/check-answers';
...
     back: checkAnswersRoute,
-    form: formContainer.form(request, document)
+    form
   };
   return h.view('nunjucks/form', view);
 };
 
+const processAction = (request, action) => {
+  const currentState = request.yar.get(SESSION_KEYS.paperFormsFlow);
+  const nextState = reducer.reducer(currentState, action);
```
Stephan-Hough, line 37: You could get rid of line 36 and 37 could change to:
`const nextState = reducer.reducer(request.pre.state, action);`

Stephan-Hough (approved):

src/internal/modules/returns-notifications/lib/controller.js
james-siteclick: I tried this and it didn't work unfortunately, I don't think the pre-handler is needed on all routes.

I have made one small tweak though which is to hide the submit button if no returns are selected.

PR #1459, opened 2020-10-28
Feat/water 2849 only 1 invoice licence per licence number per invoice

timtamimi:
Ticket: https://eaflood.atlassian.net/browse/WATER-2849

To be merged along with: https://github.com/DEFRA/water-abstraction-service/pull/1205


src/internal/modules/billing/services/transactions-csv.js
```diff
@@ -46,19 +46,6 @@ const getInvoiceAccountData = invoiceAccount => {
   };
 };
 
-const getInvoiceData = invoice => ({
```
james-siteclick, line 49: These properties appear to be coming from an `invoice` model not an `invoiceLicence`, so wondering if they need to be deleted?

Did it break without these deletions?

timtamimi: I think that while the data was mapped against an invoice model (or an invoice-like model), the data itself seemed to come from `licence_holder_address`.

Or so I assumed, because prior to removing this, the UI responded with 'An error occurred' page.

james-siteclick (approved):

PR #1461, opened 2020-10-29
Feat/water 2806 Edit charge version

Stephan-Hough:
Feat/water 2806
-- added warning to check your answers page for charge info if the new Charge Version start date = an existing Charge Version
-- new connector code to water api for get charge versions by licence id
-- logic added to helper to determine if warning should be displayed


src/internal/modules/charge-information/lib/helpers.js
```diff
@@ -85,6 +85,23 @@ const findInvoiceAccountAddress = request => {
...
+ * @param {*} draftChargeVersionStartDate the start date of the draft charge version
+ */
+const isOverridingChargeVersion = async (request, draftChargeVersionStartDate) => {
+  try {
+    const { data: chargeVersions } = await services.water.chargeVersions.getChargeVersionsByLicenceId(request.pre.licence.id);
+    return !!chargeVersions.find(version => version.dateRange.startDate === draftChargeVersionStartDate);
+  } catch (error) {
+    if (error.statusCode !== 404) {
+      return error;
+    } else { return false; }
```
DGajewska, line 100: I'm not sure this is right, we can't guarantee that there are no charge versions with the same start date if we get a 404

DGajewska (changes requested):

src/internal/modules/charge-information/lib/helpers.js
Stephan-Hough: I will change the water service to return an empty array with http 200 OK which will be a better approach but the scenario remains. An empty array will result in a false because a charge version for the licence with the same start date was not found so no warning will be displayed.

DGajewska (approved):

PR #1473, opened 2020-11-06
Ensures latest completed return cycle is selected in create bill run flow

james-siteclick:
WATER-2820

When creating a bill run, if two-part tariff is selected, the financial year is now based on the most recent completed return cycle for the selected season.


src/internal/modules/billing/controllers/create-bill-run.js
```diff
@@ -82,10 +82,8 @@ const getBatchDetails = (request, billingRegionForm) => {
     selectedTwoPartTariffSeason
   } = forms.getValues(billingRegionForm);
 
-  const financialYear = helpers.charging.getFinancialYear();
-  const financialYearEnding = selectedTwoPartTariffSeason === seasons.WINTER_AND_ALL_YEAR
-    ? financialYear - 1
-    : financialYear;
+  const isSummer = selectedTwoPartTariffSeason === seasons.SUMMER;
+  const financialYearEnding = getBatchFinancialYearEnding(selectedBillingType, isSummer);
```
Stephan-Hough, line 86: Do you not have to pass in the ref date here?

Stephan-Hough (approved):

PR #1482, opened 2020-11-13
Charge information tab fixes

james-siteclick:
* Sorts charge versions by start date then version number
* Adds the change reason text to the table
* Hide invoice account/company info for non-chargeable charge versions

### DIFF
--- a/src/internal/views/nunjucks/charge-information/view.njk
+++ b/src/internal/views/nunjucks/charge-information/view.njk
@@ -57,21 +57,24 @@ formErrorSummary %}
           {{ summaryListRow('End date', chargeVersion.dateRange.endDate | date) }}
         {% endif %}
 
-        <div class="govuk-summary-list__row">
-          <dt class="govuk-summary-list__key">Billing account</dt>
-          <dd class="govuk-summary-list__value">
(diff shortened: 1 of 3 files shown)


src/internal/views/nunjucks/charge-information/view.njk
```diff
@@ -174,23 +184,23 @@ formErrorSummary %}
   {% if isEditable %}
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-full">
-        <form method="post" action="/licences/{{licenceId}}/charge-information/check" novalidate>
+        <form method="post" action="/licences/{{licenceId}}/charge-information/check" novalidate="novalidate">
```
timtamimi, line 187: I thought `novalidate` is a boolean, so `<form .... novalidate>` should do the trick?

timtamimi (approved):

src/internal/views/nunjucks/charge-information/view.njk
james-siteclick: I think it is valid HTML5 so will leave it for now:
https://www.w3.org/TR/2008/WD-html5-20080610/semantics.html

PR #1485, opened 2020-11-17
Water-2868 - Copy existing charge data to create a new charge version

Stephan-Hough:
Feat/water 2868
-- copy existing charge data to create new charge version


src/internal/modules/charge-information/forms/use-abstraction-data.js
```diff
@@ -2,16 +2,37 @@
...
+const moment = require('moment');
 
 const { formFactory, fields } = require('shared/lib/forms/');
 const routing = require('../lib/routing');
 const { getActionUrl } = require('../lib/form-helpers');
 
+const getAddionalChoices = (chargeVersions) => {
+  const choices = [{ divider: 'or' }];
+  chargeVersions.map(cv => {
+    if (cv.status === 'current' || cv.status === 'superseded') {
```
james-siteclick, line 14: Not sure this check is necessary as the importer no longer imports draft charge versions - so all will be either current or superseded

src/internal/modules/charge-information/forms/use-abstraction-data.js
```diff
@@ -2,16 +2,37 @@
...
+      choices.push(
+        { value: cv.id,
+          label: `Use charge information valid from ${moment(cv.dateRange.startDate).format('D MMMM YYYY')}`
```
james-siteclick, line 17: Possibly a UCD question, but supposing there are multiple charge versions starting on the same date, how can the user distinguish them?

Possibly we should consider filtering to only show `current` charge versions.  This would mean there is only ever one per start date and the radio options would then be unique.  This might make sense because otherwise we are allowing the user to copy charge versions that somebody has already determined had errors.

Note after `2874` is done, there will be lots of `current` charge versions and not many `superseded`

src/internal/modules/charge-information/forms/use-abstraction-data.js
```diff
@@ -35,10 +53,12 @@ const useAbstractionDataForm = request => {
   return f;
 };
 
-const useAbstractionDataSchema = () => {
+const useAbstractionDataSchema = (request) => {
+  const { chargeVersions } = request.pre;
+  const validIds = chargeVersions.map(cv => (cv.status === 'current' || cv.status === 'superseded') ? cv.id : null);
```
james-siteclick, line 58: I guess this might need tweaking if we tweak the above

src/internal/modules/charge-information/pre-handlers.js
```diff
@@ -107,6 +107,17 @@ const loadLicencesWithWorkflowsInProgress = async request => {
   }
 };
 
+const loadChargeVersions = async request => {
+  const { licenceId } = request.params;
+  try {
+    const { data: chargeVersions } = await services.water.chargeVersions.getChargeVersionsByLicenceId(licenceId);
+    return chargeVersions.filter(cv => (cv.status === 'current' || cv.status === 'superseded'))
+      .sort((rowA, rowB) => new Date(rowB.dateRange.startDate) - new Date(rowA.dateRange.startDate));
```
james-siteclick, line 115: Looks like charge versions are already filtered in above code, so filtering may not be needed here.
I generally sort by start date then version number, so this might become:

```
const { sortBy } = require('lodash');
// ...
return sortBy(chargeVersions, ['dateRange.startDate', 'versionNumber']);
```

james-siteclick (changes requested):

src/internal/modules/charge-information/forms/use-abstraction-data.js
james-siteclick: I think the filtering of the charge versions has gone from above so maybe not needed here either?

james-siteclick (changes requested):
Couple of minor tweaks

src/internal/modules/charge-information/forms/use-abstraction-data.js
Stephan-Hough: of course thanks

src/internal/modules/charge-information/forms/use-abstraction-data.js
```diff
@@ -35,10 +53,12 @@ const useAbstractionDataForm = request => {
   return f;
 };
 
-const useAbstractionDataSchema = () => {
+const useAbstractionDataSchema = (request) => {
+  const { chargeVersions } = request.pre;
+  const validIds = filterChargeVersions(chargeVersions).map(cv => (cv.status === 'current') ? cv.id : null);
```
james-siteclick, line 58: I think the filter already ensures there are only current ones - could this line be simplified to:
```
const validIds = filterChargeVersions(chargeVersions).map(cv => cv.id);
```

james-siteclick (changes requested):

james-siteclick (approved):

PR #1502, opened 2020-12-01
Bugfix: Limit volumes to 6 decimal places

DGajewska:
WATER-2937

* Adds hint text to billing quantity in TPT review
* Adds hint to quantities page in charge info flow
* Updates suffix mapping to use new feature in textInput macro

### DIFF
--- a/src/internal/modules/charge-information/forms/charge-element/quantities.js
+++ b/src/internal/modules/charge-information/forms/charge-element/quantities.js
@@ -6,24 +6,28 @@ const { capitalize } = require('lodash');
 const { CHARGE_ELEMENT_STEPS } = require('../../lib/charge-elements/constants');
 const { getChargeElementData, getChargeElementActionUrl } = require('../../lib/form-helpers');
 
+const getErrors = key => {
+  const errors = { 'string.regex.base': {
+    message: `Enter a number for the ${key} quantity using 6 decimal places or fewer, the number must be more than 0`
+  } };
+  if (key === 'authorised') {
+    const requiredAuthorisedQuantityError = {
+      message: `Enter an authorised quantity`
+    };
+    errors['any.required'] = requiredAuthorisedQuantityError;
+    errors['any.empty'] = requiredAuthorisedQuantityError;
+  };
+
+  return errors;
+};
+
 const getFormField = (key, data) => {
   const fieldName = `${key}AnnualQuantity`;
   return fields.text(fieldName, {
     controlClass: 'govuk-input govuk-input--width-10',
     label: `${capitalize(key)}${key === 'billable' ? ' (optional)' : ''}`,
-    type: 'number',
     suffix: 'megalitres per year',
-    errors: {
-      'any.required': {
-        message: `Enter an ${key} quantity`
-      },
-      'any.empty': {
-        message: `Enter an ${key} quantity`
-      },
-      'string.regex.base': {
-        message: `Enter a number for the ${key} quantity using 6 decimal places or fewer, the number must be more than 0`
-      }
-    }
+    errors: getErrors(key)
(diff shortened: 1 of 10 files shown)


src/internal/modules/charge-information/forms/charge-element/quantities.js
```diff
@@ -11,17 +11,17 @@ const getFormField = (key, data) => {
   return fields.text(fieldName, {
     controlClass: 'govuk-input govuk-input--width-10',
     label: `${capitalize(key)}${key === 'billable' ? ' (optional)' : ''}`,
+    type: 'number',
     suffix: 'megalitres per year',
-    mapper: 'numberMapper',
     errors: {
-      'any.empty': {
+      'any.required': {
         message: `Enter an ${key} quantity`
```
timtamimi, line 18: Erm. Really silly thing: Could this end up rendering `Enter **an valid** billable quantity`? If so, could we do something to fix it so that it's not grammatically incorrect?

(I appreciate this is pedantry on my part)

src/internal/modules/charge-information/forms/charge-element/quantities.js
```diff
@@ -11,17 +11,17 @@ const getFormField = (key, data) => {
...
       },
-      'number.base': {
-        message: `Enter a valid ${key} quantity as a number that is more than zero`
+      'any.empty': {
+        message: `Enter an ${key} quantity`
       },
-      'number.greater': {
-        message: `Enter a valid ${key} quantity as a number that is more than zero`
+      'string.regex.base': {
+        message: `Enter a number for the ${key} quantity using 6 decimal places or fewer, the number must be more than 0`
```
timtamimi, line 24: May I suggest instead of `the number must be more than 0` saying `the quantity must be greater than 0`?

Again, pedantry. So feel free to ignore this

test/internal/modules/charge-information/controllers/charge-element.js
```diff
@@ -100,8 +100,8 @@ const validPayload = {
     'endDate-month': '10'
   },
   quantities: {
-    authorisedAnnualQuantity: 1234,
-    billableAnnualQuantity: null
+    authorisedAnnualQuantity: '1234',
+    billableAnnualQuantity: ''
```
timtamimi, line 104: No drama but I think you have since made this nullable so feel free to swap out `''` for `null` (Or don't - It doesn't matter)

Or better still: Feel free to have tests for both null and `''`

timtamimi (changes requested):
Few text suggestions. Feel free to ignore those if you don't feel they need re-adjusting. Just let me know once you've looked at them so I can give this a ✅

test/internal/modules/charge-information/controllers/charge-element.js
DGajewska: I changed this from a number to a string so that I could validate it with a regex, so I don't think that it will ever be a null

src/internal/modules/charge-information/forms/charge-element/quantities.js
DGajewska: This came out of a chat with UCD, so I'd rather not change it

src/internal/modules/charge-information/forms/charge-element/quantities.js
DGajewska: I was going to fix this, but then I realised that this might be a bit misleading as the `'any.required'` and `'any.empty'` errors will only ever show up for the authorised quantity because the billable quantity is optional. I'll try to make it a bit clearer

timtamimi: Cool. Thank you :)

timtamimi (approved):

PR #1510, opened 2020-12-09
Feat: Contact Entry plugin

DGajewska:
WATER-2963

* Renames existing 'contact-entry' module to 'account-entry' to match new naming convention
* Updates invoice accounts module to work with contact plugin
* Moves existing contact entry routes to own module
* Creates configurable plugin
* Moves shared pre-handlers to shared library


Stephan-Hough (approved):

src/internal/modules/contact-entry/plugin.js
```diff
@@ -1,10 +1,51 @@
...
+function getNewContact (key) {
+  return (session.get(this, key) || {}).data;
+}
 
 const contactEntryPlugin = {
-  register: (server) => {
-    for (const eachRoute of Object.values(routes)) {
-      server.route(eachRoute);
-    }
+  register: (server, options) => {
```
Stephan-Hough, line 42: do you need the options here as per sonar cloud?

DGajewska: It is done :)

PR #1524, opened 2020-12-16
Adds external user registration flow to acceptance test suite

james-siteclick:


test/regression/external/helpers/acceptance-tests-proxy.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
```
timtamimi, line 1: You don't need this file (I'll explain below in the other comment...)

test/regression/external/user-registration.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    expect(browser).toHaveUrlContaining('/add-licences');
+    expect(getPageTitle()).toHaveText('Add your licences to the service');
+  });
+
+  /**
+   * @todo flow where user enters email twice
+   */
+
+  after(async () => {
+    await tearDown();
```
timtamimi, line 139: Instead of calling this tearDown() function which in turns calls a REST endpoint, you may find it easier/more maintainable to use the built-in Wdio event listeners like [this one here](https://github.com/DEFRA/water-abstraction-ui/blob/89701a81376e75ed17dded98a660e848d3664426/wdio.conf.js#L270) alongside the setup routine that happens [here](https://github.com/DEFRA/water-abstraction-ui/blob/89701a81376e75ed17dded98a660e848d3664426/wdio.conf.js#L167).

timtamimi (changes requested):

timtamimi (approved):

PR #1527, opened 2020-12-22
Reg-Pack: Resetting password as internal and external user

DGajewska:


test/regression/external/reset-password.spec.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  it('shows a link to resend the reset password email', () => {
+    expect($('a[href="/reset_password_resend_email"]')).toHaveText('Has the email not arrived?');
+  });
+
+  it('clicks the link in the confirmation email', () => {
+    const resetUrl = getPersonalisation(config.baseUrl, EMAIL_ADDRESS, 'reset_url');
+    browser.url(resetUrl);
+    expect(browser).toHaveUrlContaining('/reset_password_change_password?');
```
Stephan-Hough, line 58: I think you need to remove the '?' as this does not appear in the URL

timtamimi (approved):

timtamimi (changes requested):
Sorry, just spotted that this is calling the setup endpoint in `before`. This needs to come out, as WDIO is configured to globally call the setup endpoint when the testing starts. Same goes for the teardown routine.

timtamimi (approved):
I think this can be merged now. Thank you :)

PR #1528, opened 2020-12-22
Bugfix/water-2956 - Paper forms - licences with no returns due

Stephan-Hough:
bug/water-2956
-- new banner included in views when licence with no returns due included
-- user routed back to enter licences when all licences has no returns
-- tests updated


src/internal/modules/returns-notifications/controllers/paper-forms.js
```diff
@@ -37,14 +38,18 @@ const roleMapper = require('shared/lib/mappers/role');
  * they wish to send paper return forms
  */
 const getEnterLicenceNumber = async (request, h) => {
-  return h.view('nunjucks/form', {
+  const { licencesWithNoReturns } = request.query;
```
timtamimi, line 41: Is storing this in request.query the best option? That URL can get quite long...

Could we pop it in the session with Yar or perhaps load it in request.pre?

src/internal/modules/returns-notifications/controllers/paper-forms.js
```diff
@@ -111,10 +121,11 @@ const getRecipientCount = state => Object.values(state)
  */
 const getCheckAnswers = async (request, h) => {
   const isRecipients = getRecipientCount(request.pre.state) > 0;
-
+  console.log(licencesWithNoReturnsDue(request.pre.state));
```
timtamimi, line 124: You may have accidentally left this in.

src/internal/modules/returns-notifications/lib/reducer.js
```diff
@@ -77,6 +77,9 @@ const mapLicencesToState = (licences, refDate) => {
     licenceRow.documents.forEach(documentRow =>
       acc.set(documentRow.document.id, mapDocumentRow(licenceRow, documentRow, refDate))
     );
+    if (licenceRow.documents.length === 0) {
```
timtamimi, line 80: This is interesting. Does this cater for an edge case? If so, it might be worth putting an inline comment explaining why? (It's not a problem - just curious)

timtamimi (changes requested):

timtamimi (approved):

PR #1534, opened 2021-01-11
KPI + Service Status page rejig

timtamimi:
Original ticket: https://eaflood.atlassian.net/browse/WATER-2978

What happened here:
- Moved Service Status page from 'shared' to 'internal' only.
- Removed the CRM/IDM KPIs from the view of the Service Status page.
- Changed the base route for the KPI page to /reporting/*
- Moved the KPI page from 'shared' to 'internal' only
- Added auth configuration to the KPI page, such that you can only access it if you can see the manage tab.


DGajewska (changes requested):
Just a small comment to remove the empty file `test/external/modules/kpi-reporting/routes.js`. Otherwise, it's good to go :)

DGajewska (approved):

PR #1547, opened 2021-01-29
Bugfix/water 2996 - Minor UI fix for Charge Version Workflow

timtamimi:
Original ticket: https://eaflood.atlassian.net/browse/WATER-2996

### DIFF
--- a/src/internal/modules/charge-information/lib/helpers.js
+++ b/src/internal/modules/charge-information/lib/helpers.js
@@ -85,8 +85,8 @@ const getLicencePageUrl = async licence => {
 
 const isCurrentAddress = invoiceAccountAddress => invoiceAccountAddress.dateRange.endDate === null;
 
-const getCurrentBillingAccountAddress = billingAccount => billingAccount.invoiceAccountAddresses
-  .find(isCurrentAddress);
+const getCurrentBillingAccountAddress = billingAccount => billingAccount && billingAccount.invoiceAccountAddresses ? billingAccount.invoiceAccountAddresses
+  .find(isCurrentAddress) : null;
 
 /**
  * Checks if the new draft charge version has the same start date as an existing charge version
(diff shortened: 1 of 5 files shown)


src/internal/modules/charge-information/lib/helpers.js
```diff
@@ -85,8 +85,8 @@ const getLicencePageUrl = async licence => {
 
 const isCurrentAddress = invoiceAccountAddress => invoiceAccountAddress.dateRange.endDate === null;
 
-const getCurrentBillingAccountAddress = billingAccount => billingAccount.invoiceAccountAddresses
-  .find(isCurrentAddress);
+const getCurrentBillingAccountAddress = billingAccount => billingAccount && billingAccount.invoiceAccountAddresses ? billingAccount.invoiceAccountAddresses
+  .find(isCurrentAddress) : undefined;
```
james-siteclick, line 90: This could maybe be simplified a bit with lodash `get`
```
const getCurrentBillingAccountAddress = billingAccount => get(billingAccount, 'invoiceAccountAddresses', []).find(isCurrentAddress);
```
I think as a bonus it will increase the test coverage by removing the ternary operator.

test/internal/modules/charge-information/lib/helpers.js
```diff
@@ -104,6 +104,25 @@ experiment('internal/modules/charge-information/lib/helpers', () => {
     });
   });
 
+  experiment('.getCurrentBillingAccountAddress', () => {
+    const address1 = { id: 'address1', dateRange: { endDate: null } };
+    const address2 = { id: 'address2', dateRange: { endDate: new Date() } };
```
james-siteclick, line 109: Not that I think it affects the test outcome, but I think the `endDate` here would be a string in the form YYYY-MM-DD

james-siteclick (changes requested):

james-siteclick (approved):

PR #1570, opened 2021-02-08
Alternative regex

timtamimi:
Modifies the regex to allow for values starting with zero, as long as they're neither negative nor zero. E.g. 0.123 or 0.6

Relates to https://eaflood.atlassian.net/browse/WATER-3014

### DIFF
--- a/src/internal/modules/charge-information/forms/charge-element/quantities.js
+++ b/src/internal/modules/charge-information/forms/charge-element/quantities.js
@@ -58,7 +58,7 @@ const form = request => {
 };
 
 const schema = (request) => {
-  const nonZeroNumberWithWSixDpRegex = /^[1-9]\d*\.?\d{0,6}$/;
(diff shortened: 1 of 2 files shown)


src/internal/modules/charge-information/forms/charge-element/quantities.js
```diff
@@ -58,7 +58,7 @@ const form = request => {
 };
 
 const schema = (request) => {
-  const nonZeroNumberWithWSixDpRegex = /^[1-9]\d*\.?\d{0,6}$/;
+  const nonZeroNumberWithWSixDpRegex = /^\s*(?=.*[1-9])\d*(?:\.\d{1,6})?\s*$/;
```
james-siteclick, line 61: I don't 100% follow the regex but I have tested it and it appears to fix the issues described.
We could consider removing the `\s*` here, and adding `.trim()` in the Joi validator before the regex.  This should trim the whitespace before it reaches the regex.

james-siteclick (approved):

PR #1592, opened 2021-02-19
WATER-3016 store CM data/transactions in DB

james-siteclick:
`WATER-3016`

Tweaks UI to accept new shape of data from water service for charging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "water-abstraction-ui",
-  "version": "1.61.0",
+  "version": "1.60.0",
   "lockfileVersion": 1,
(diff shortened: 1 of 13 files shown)


timtamimi (approved):
I've left a suggestion to add an inline comment which could be addressed or discarded as you see fit. Happy for this to get merged as-is.

timtamimi (changes requested):
~CORRECTION: Marking this as 'changes requested' because of the discrepancy between `develop` and this branch, particularly in `src/internal/lib/connectors/services/water/index.js`~

timtamimi (approved):

PR #1624, opened 2021-03-11
Water-3020 external user returns regression tests

Stephan-Hough:
Feat/water 3020
-- regression tests added for the external user returns submission


test/regression/external/submit-returns.spec.js
```diff
@@ -0,0 +1,440 @@
...
+    expect($('form')).toBeVisible();
+    const formQuestion = $('//form/p');
+    expect(formQuestion).toHaveTextContaining('Enter your readings exactly as they appear on your meter');
+    const button = getButton();
+    button.click();
+  });
+
+  checkErrorMessage('Enter a meter start reading');
+
+  it('selects cubic meters', () => {
```
james-siteclick, line 180: There is additional logic here it would be good to add tests for:

* Initial reading cannot be negative
* Meter readings must not be negative
* Meter readings must keep increasing (e.g. March reading cannot be lower than Feb)

It would be good to include tests for:
* All fields left empty
* Some fields left empty
* Non-numeric values
* Zero values
* Negative values
* Volumes entered in each different unit available - are they calculated correctly

test/regression/external/submit-returns.spec.js
```diff
@@ -0,0 +1,440 @@
...
+    expect($('form')).toBeVisible();
+    const formQuestion = $('//form/div/fieldset/legend');
+    expect(formQuestion).toHaveTextContaining('Which units are you using?');
+    const radioButton2 = $('[value="l"]');
+    radioButton2.click();
+    const button = getButton();
+    button.click();
+  });
+
+  it('enters the meter readings', () => {
```
james-siteclick, line 380: This should probably say 
```
it('enters the volumes')
```
Or similar as these are not meter readings in this flow

test/regression/external/submit-returns.spec.js
```diff
@@ -0,0 +1,440 @@
...
+    expect($('form')).toBeVisible();
+    const formQuestion = $('//form/p');
+    expect(formQuestion).toHaveTextContaining('Your abstraction volumes');
+    const textBoxJan = $('input[name="2020-01-01_2020-01-31"]');
+    textBoxJan.setValue('10000');
```
james-siteclick, line 385: It be good if it tested the logic on this page a bit more, e.g what happens if:

* All fields left empty
* Some fields left empty
* Non-numeric values
* Zero values
* Negative values
* Volumes entered in each different unit available - are they calculated correctly

test/regression/external/submit-returns.spec.js
```diff
@@ -0,0 +1,440 @@
...
+    textBoxOct.setValue('10000');
+    const textBoxNov = $('input[name="2020-11-01_2020-11-30"]');
+    textBoxNov.setValue('10000');
+    const textBoxDec = $('input[name="2020-12-01_2020-12-31"]');
+    textBoxDec.setValue('10000');
+    const button = getButton();
+    button.click();
+  });
+
+  it('enters the meter readings', () => {
```
james-siteclick, line 412: 'enters the meter details'

james-siteclick (changes requested):
I had to run
`npm i chromedriver@89 --save-dev` 
to get this to work.

I think there are a few different paths through this flow that are not exercised -e.g. if you do estimated volumes.  Maybe we should include tests for these too.

I'm not certain why it's necessary to set up multiple licences for this test - a single licence can have many returns.

test/regression/external/submit-returns.spec.js
Stephan-Hough: done

test/regression/external/submit-returns.spec.js
Stephan-Hough: Added tests for the above except for non-numerics because the UI does not accept non-numeric values

test/regression/external/submit-returns.spec.js
Stephan-Hough: Additional tests added as requested

test/regression/external/submit-returns.spec.js
Stephan-Hough: done

Stephan-Hough:
> I had to run
> `npm i chromedriver@89 --save-dev`
> to get this to work.
> 
> I think there are a few different paths through this flow that are not exercised -e.g. if you do estimated volumes. Maybe we should include tests for these too.
> 
> I'm not certain why it's necessary to set up multiple licences for this test - a single licence can have many returns.

All the paths have now been included in this work package except for the estimated volumes. I assume this was left out of the scope of this ticket due to the overlap of logic for submitting estimated and actual metered returns. 

Four licences are set up as the barebones for test data. I am not entirely clear on what the concern may be with this but this has been the case for the previous acceptance tests suite as well. Perhaps the scope of the tickets could have been refined more precisely on what data should have been used.

james-siteclick (approved):

PR #1645, opened 2021-03-19
Fix: Updates bills tab

DGajewska:
WATER-3042

* Adds link to 'All sent bills' page if more than 10 bills to view
* Fixes table caption to use correct formatting
* Changes allowed scope on route definition and removes isChargingUser flag in controller and template
* Adds Bills tab to regression pack


src/internal/modules/licences/routes/bills-tab.js
```diff
@@ -4,7 +4,7 @@ const Joi = require('@hapi/joi');
 const controller = require('../controllers/bills-tab');
 const { scope } = require('../../../lib/constants');
 
-const allowedScopes = [scope.billing];
+const allowedScopes = [scope.charging];
```
james-siteclick, line 7: I seem to have both a "charging" and "billing" role in the IDM.

I think we need to ensure all services are using the same one and delete the other - although that's not necessarily within the scope of this ticket.

src/internal/views/nunjucks/billing/macros/invoices-table.njk
```diff
@@ -183,10 +183,10 @@
     </td>
   {% endmacro %}
 
-  {% macro invoiceLink(invoice) %}
+  {% macro invoiceLink(row) %}
     <td class="govuk-table__cell" scope="row">
-      <a href="/billing/batch/{{ invoice.billingBatch.id }}/invoice/{{ invoice.id }}">
-        {{invoice.invoiceNumber}}
+      <a href="/billing/batch/{{ row.batch.id }}/invoice/{{ row.invoice.id }}">
+        {{row.invoice.invoiceNumber}}
```
james-siteclick, line 189: Does this need to cope with the `invoiceNumber` being blank in some cases?

james-siteclick (approved):

PR #1667, opened 2021-04-06
Implements new GOV.uk cookie banner

james-siteclick:
* Uses new design of GOV.uk cookie banner
* Allows users to opt in/out of analytics cookies
* Updates cookie page content including form for opt in/out
* Tweaks Catbox/Redis settings to use lazy connect in unit tests
* Removes/rewrites broken legacy unit tests

**Note**: run `npm i && npm run install-assets`


src/shared/plugins/cookie-message/routes.js
```diff
@@ -0,0 +1,47 @@
...
+    handler: controller.getSetCookiePreferences,
+    config: {
+      auth: {
+        strategy: 'standard',
+        mode: 'try'
+      },
+      description: 'View cookie details and set choices',
+      validate: {
+        query: Joi.object({
+          acceptAnalytics: Joi.boolean().truthy('1').falsy('0'),
```
timtamimi, line 41: Just purely out of interest, can't we just take actual booleans here? E.g. `true` or `false` rather than `1` vs `0`?

src/shared/plugins/cookie-message/index.js
```diff
@@ -1,42 +1,162 @@
...
+/**
+ * Gets the options for declaring the cookie that will manage the
+ * preferences
+ *
+ * @returns {Object}
+ */
+const getCookieOptions = () => ({
+  isSecure: process.env.NODE_ENV !== 'local',
+  isHttpOnly: true,
+  ttl: 365 * 24 * 60 * 60 * 1000,
```
timtamimi, line 127: Should this be 28 days rather than 365?

timtamimi (approved):
Looks good! I added two optional queries.

timtamimi (changes requested):
I think we need to update the content that appears on /cookies on the external site such that we make it clear that the session cookie is 2 hours (Currently shown as 10 years?)

src/shared/plugins/cookie-message/routes.js
james-siteclick: As they are decoded query params they will always be strings.  But we could do:
```
Joi.boolean().truthy('true').falsy('false'),
```
And then the URL would be /blah?acceptAnalytics=true
```
Do you think that is neater?

src/shared/plugins/cookie-message/index.js
james-siteclick: I looked at some other GOV.uk sites and they seemed to set it for a year, otherwise the banner will pop up monthly

timtamimi: Cool. If that's the done thing then let's leave it 👍🏽

timtamimi (approved):

PR #1668, opened 2021-04-06
Fixes two broken unit tests

timtamimi:
~This PR is into the release branch, because the release branch will be merged into Develop!~

I have changed base into `develop` since it looks unlikely that this will be resolved today.


test/internal/modules/agreements/lib/reducer.js
```diff
@@ -49,7 +50,7 @@ experiment('internal/modules/agreements/lib/reducer', () => {
         code: 'S127',
         dateSigned: undefined,
         isDateSignedKnown: false,
-        startDate: '2020-04-01'
+        startDate: `${moment().year()}-04-01`
```
james-siteclick, line 53: I'm not certain this will work in some cases - e.g. if today's date is in Jan/Feb/March.

The root of this issue looks like the reducer is not a pure function as it is supposed to be - I think the solution would be to pass the current date as an argument to the action creator.

test/internal/modules/billing/controllers/create-bill-run.js
```diff
@@ -416,8 +417,8 @@ experiment('internal/modules/billing/controllers/create-bill-run', () => {
         await controller.postBillingBatchRegion(request, h);
 
         const [batch] = services.water.billingBatches.createBillingBatch.lastCall.args;
-        const financialYear = helpers.charging.getFinancialYear(new Date());
-        expect(batch.financialYearEnding).to.equal(financialYear - 1);
+        const financialYear = helpers.charging.getFinancialYear(moment(billingRegions.data[0].dateCreated).toDate());
```
james-siteclick, line 420: I'm struggling to see why there would be a relationship between the region creation date and the financial year of the billing batch here.

I think for TPT winter it should run the financial year of the last winter/all year return cycle (although I would have to check this - it may also take into account allowing for the due date + grace period).   

So stubbing the `getReturnCycles` function in the helpers may be a way to get a consistent financial year here.

Either that or possibly something like this which would check the month:
```
const financialYear = moment().month() > 2 ? moment().year() : moment().year()-1;
```

james-siteclick (changes requested):

test/internal/modules/billing/controllers/create-bill-run.js
timtamimi: I thought about stubbing `getReturnCycles` but decided against it purely because I wasn't sure what the expected outcome from that function should be. 

For that reason, I've gone ahead and implemented the second suggestion, where financial year is calculated on the fly. The only difference is that I did `moment().month() > 3` instead of `moment().month() > 2` - Let me know if that's incorrect!

test/internal/modules/agreements/lib/reducer.js
timtamimi: I did something slightly off-script: I have left the dates array the same, but I have made a change such that the financial year is calculated based on today's date, if the date signed is not provided. Curious if you think that's OK.

If not, I can add today's date into the `dates` array - the reason I opted not to do that is because I think it'll break other tests if I do...

timtamimi:
Superseded by https://github.com/DEFRA/water-abstraction-ui/pull/1670

PR #1672, opened 2021-04-08
Feat/water 2864

timtamimi:
Service PR: https://github.com/DEFRA/water-abstraction-service/pull/1353
UI PR: https://github.com/DEFRA/water-abstraction-ui/pull/1672
Reporting PR: https://github.com/DEFRA/water-abstraction-reporting/pull/4

Main ticket: https://eaflood.atlassian.net/browse/WATER-2864


src/internal/modules/reporting/controllers.js
```diff
@@ -0,0 +1,22 @@
...
+const getChargingForecastReportsPage = (request, h) => {
+  return h.view('nunjucks/reporting/charging-forecast-reports', {
+    ...request.view,
+    back: '/manage',
+    pageTitle: 'Download a charging forecast report'
+  });
+};
+
+const getDownloadableReport = async (request, h) => {
+  const { reportIdentifier } = request.params;
```
james-siteclick, line 12: If the water service is responding with a CSV file, we should be able to pipe it directly through.

james-siteclick (changes requested):

src/internal/modules/reporting/controllers.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Loads a specific report by making a request to the backend
+ * @param request {Object} The request object
+ * @param h {Object} The response object
+ * @returns {Object} h.response Response body
+ */
+const getDownloadableReport = async (request, h) => {
+  const { reportIdentifier } = request.params;
+  // get signed url
+  const report = await services.water.reporting.getReport(request, h);
```
james-siteclick, line 26: Although I don't think it's causing an issue, I don't think the `await` is needed here

src/internal/lib/connectors/services/water/ReportingService.js
```diff
@@ -0,0 +1,24 @@
+const ServiceClient = require('shared/lib/connectors/services/ServiceClient');
+const got = require('got');
+
+class ReportingService extends ServiceClient {
+  /**
+   * Requests a report from the API
+   * @param request {Object} The request object
+   * @returns {Request} Request body
+   */
+  getReport (request) {
```
james-siteclick, line 10: I think I would expect only the needed params to be passed in here, e.g. maybe:
```
getReport (userId, reportIdentifier)
```
This would make the connector code more flexible in case it's needed elsewhere

james-siteclick (changes requested):
It's cool that you got this working with the `got` library - nice neat way of implementing the streams.

Couple of tiny suggested tweaks

src/internal/modules/reporting/controllers.js
timtamimi: I THINK it's needed - I found from experimenting with it that if I omit it, only the first chunk is downloaded 😁

src/internal/lib/connectors/services/water/ReportingService.js
timtamimi: Sure thing. This is done in 5deb30d1

james-siteclick (approved):

PR #1684, opened 2021-04-16
WATER-2994: Return additional factor fields on transactions csv

philhunt-defra:
We now return the factors used in calculations so they can be provided for audit purposes. The 'S130 agreement value' field has been left intentionally blank for future use.

### DIFF
--- a/src/internal/modules/billing/services/transactions-csv.js
+++ b/src/internal/modules/billing/services/transactions-csv.js
@@ -13,17 +13,26 @@ const getAbsStartAndEnd = absPeriod => ({
   'Abstraction period end date': moment().month(absPeriod.endMonth - 1).date(absPeriod.endDay).format('D MMM')
 });
 
-const getAgreementsString = agreements => agreements.map(agreement => agreement.code).join(', ');
+const findAgreementValue = (agreements, code) => {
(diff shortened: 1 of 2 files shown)


src/internal/modules/billing/services/transactions-csv.js
```diff
@@ -13,17 +13,26 @@ const getAbsStartAndEnd = absPeriod => ({
   'Abstraction period end date': moment().month(absPeriod.endMonth - 1).date(absPeriod.endDay).format('D MMM')
 });
 
-const getAgreementsString = agreements => agreements.map(agreement => agreement.code).join(', ');
+const findAgreementValue = (agreements, code) => {
+  const value = (agreements || []).filter(agreement => agreement.code.indexOf(code) > -1);
+  return value && value.length > 0 ? `S${value[0].code}` : null;
```
timtamimi, line 18: Where you return ``S${value[0].code}``, you can return ``${value[0].code}``

This is because the 'S' is part of the agreement code already, so currently this will produce `SS130x`

timtamimi (changes requested):

timtamimi (approved):

PR #1707, opened 2021-04-29
Tweak validation

james-siteclick:
WATER-3155


src/shared/plugins/cookie-message/routes.js
```diff
@@ -4,6 +4,12 @@ const Joi = require('@hapi/joi');
 const { createRoutePair } = require('shared/lib/route-helpers');
 const controller = require('./controller');
 
+const redirectPath = Joi
+  .string()
+  .regex(/^\/[^/].*/)
```
timtamimi, line 9: Just out of interest - Is this regex necessary? Wondering if `.uri({ relativeOnly: true })` is sufficient.

timtamimi (approved):
Approved but left a comment

PR #1709, opened 2021-04-29
Regression tests/user creation - Internal

srujan-chada:


cypress/integration/internal/create-user.spec.js
```diff
@@ -0,0 +1,56 @@
...
+      cy.get('.govuk-button').should('contain.text', 'Continue');
+    });
+    describe('populates the email field and submits the form', () => {
+      let tempEmail = `regression.tests.${uuid()}@defra.gov.uk`;
+      cy.get('input#email').type(tempEmail);
+      cy.get('form > .govuk-button').click();
+    });
+
+    describe('verify the contents on the page', () => {
+      // expect($('form[action="/account/create-user/set-permissions"]')).toBeVisible();
```
timtamimi, line 44: Why is this line of commented-out code here?

timtamimi (changes requested):

timtamimi (approved):
Looks fine. Is it worth adding a fail test or two?

E.g. Trying to create a user where the email is already in use

E.g. Trying to add a user with an invalid email address.

E.g. Trying to add a user with an account that shouldn't have permission to create users

PR #1711, opened 2021-04-30
WATER-2569: Throw an error if the user does not select a file to upload

philhunt-defra:
In bulk returns file upload page, uploading no file will cause the API to get through the full file process (creating an empty file, virus scanning, etc) and then return an invalid file message.

This change checks the user has actually uploaded a file first, and then returns the appropriate error if they haven't, without running through the whole file upload function.

### DIFF
--- a/src/external/modules/returns/lib/upload-helpers.js
+++ b/src/external/modules/returns/lib/upload-helpers.js
@@ -21,7 +21,7 @@ const getErrorMessage = key => {
     'invalid-xml': 'The selected file must use the template',
     'invalid-csv': 'The selected file must use the template',
     'invalid-type': 'The selected file must be a CSV or XML file',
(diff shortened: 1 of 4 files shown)


src/external/modules/returns/lib/upload-helpers.js
```diff
@@ -21,6 +21,7 @@ const getErrorMessage = key => {
     'invalid-xml': 'The selected file must use the template',
     'invalid-csv': 'The selected file must use the template',
     'invalid-type': 'The selected file must be a CSV or XML file',
+    'no-file': 'Please select a CSV or XML file',
```
james-siteclick, line 24: Looking at the design system (https://design-system.service.gov.uk/components/file-upload/) I wonder if the message here should be:
"Select a CSV or XML file"

james-siteclick (approved):

PR #1719, opened 2021-05-06
Allow charge versions to be edited

philhunt-defra:
This change adds the capability for a Billing & Data user to make tweaks to a charge version while it is in the ‘Review’ state as part of the Charge Information Workflow.


Stephan-Hough (changes requested):
The ticket requests a billing & data (B&D) user to make changes whilst reviewing a draft charge version i.e. a B&D user should be able to click on a change link for any part of the draft charge version, make the change and then land back on the review screen to approve the charge version. 

At the moment after a change has been made the B&D user can only resubmit the charge version for review. This means the user has to go back to the charge version workflow or licence page to then come back to the review page. This unnecessarily elongates the process.

Stephan-Hough (approved):

james-siteclick (approved):

PR #1733, opened 2021-05-10
Destroy the session if the user account is disabled

philhunt-defra:
When we have active session for an internal user and if we delete the user then the session is still active and user can perform all actions.

This change checks to see if the user account has been disabled and invalidates the session if it has.


james-siteclick (changes requested):
Code looks good, but CI seems to be failing
(maybe try `npm test && npm run lint`)

PR #1745, opened 2021-05-14
Initial work to move licence summary page to use licence ID

james-siteclick:
`WATER-2880`

* Removes dependence on licence-data and view-licence plugins for internal licence summary page
* Implements page using controller and pre-handler approach 
* Uses licence ID
* Updates paths and removes pre-handlers from routes which needed to look up the licence ID by document ID
* Removes "Expired licence" page as new licence summary page works for expired/future dated licences


src/internal/views/nunjucks/internal-search/index.njk
```diff
@@ -53,7 +53,7 @@
 {% endmacro %}
 
 {% macro documentLink(doc) %}
```
timtamimi, line 55: Is it worth renaming the macro to `licenceLink`?

src/internal/views/nunjucks/internal-search/index.njk
```diff
@@ -53,7 +53,7 @@
...
-  <a href="/{{ "expired-" if not doc.isCurrent }}licences/{{ doc.documentId }}">
+  <a href="/licences/{{ doc.licence.id }}">
```
timtamimi, line 56: I would suggest wrapping this <a> ref in a conditional that checks for the licence.id being "there" first. E.g.

```
{% if doc.licence.id %}
<a...... >{{ doc.licenceNumber}}</a>
{% elseif %}
{{ doc.licenceNumber}}
{% endif %}
```

Purely because when I was poking around locally, I did find a few random results where the licence ID wasn't there and as a result, tapping on the link just took me to /licences

src/internal/views/nunjucks/view-licences/tabs/charge.njk
```diff
@@ -165,6 +165,8 @@
   <p>No agreements for this licence.</p>
 {% endif %}
 
-<a href="/licences/{{ licenceId }}/agreements/select-type" role="button" draggable="false" class="govuk-button govuk-button--secondary" data-module="govuk-button">
-  Set up a new agreement
-</a>
+{{ govukButton({
```
timtamimi, line 168: Worth wrapping this in a conditional so that we don't allow the user to create an agreement for an inactive licence?

Same question for the Charge Version creation buttons above in the same file (charge.njk), do we want to wrap those button in a conditional?

Without a conditional around those buttons, I currently *can* create a charge version for a "dead licence". I assume that's not the desired effect, but I'm not too sure.

timtamimi (approved):
I left a few "advisory comments" which you may or may not want to look at, but I'm happy with this all in all so I'm cool with it if you want to merge it as is. So I've marked it as Approved.

src/internal/views/nunjucks/view-licences/tabs/charge.njk
james-siteclick: I think this behaviour is OK.  E.g. they may need to add charge versions/agreements to a licence that has not yet started, or correct a problem with one that has recently expired

src/internal/views/nunjucks/internal-search/index.njk
james-siteclick: Done

src/internal/views/nunjucks/internal-search/index.njk
james-siteclick: Done

PR #1775, opened 2021-05-26
Show not given when abstraction period is null

philhunt-defra:
Display 'Not Given' when an abstraction period does not exist


src/shared/view/nunjucks/macros/returns/return-header.njk
```diff
@@ -114,11 +114,15 @@
...
         <dd class="meta__value">
-          {{ return.metadata.nald.periodStartDay }}
-          {{ return.metadata.nald.periodStartMonth | date('MMMM') }}
-          to
-          {{ return.metadata.nald.periodEndDay }}
-          {{ return.metadata.nald.periodEndMonth | date('MMMM') }}
+          {% if (return.metadata.nald.periodStartDay == "null" or return.metadata.nald.periodStartMonth == "null") %}
+            Not given
+          {% else %}
+            {{ return.metadata.nald.periodStartDay }}
```
james-siteclick, line 120: We do have an existing `abstractionPeriod` Nunjucks filter which encapsulates some of this logic.  You may be able to do:
```
{% set absPeriod = {
  startDay: return.metadata.nald.periodStartDay,
  startMonth: return.metadata.nald.periodStartMonth,
  endDay:  return.metadata.nald.periodEndDay,
  endMonth: return.metadata.nald.periodEndMonth
} %}

{{ absPeriod | abstractionPeriod }}

james-siteclick (approved):

src/shared/view/nunjucks/macros/returns/return-header.njk
philhunt-defra: I tried this but it fails with `Invalid abstraction period`. An alternative is to update the abstraction-period.js file to handle null values and return 'Not given', what do you think?

PR #1795, opened 2021-06-07
Feat/water 3190 remove invoice licence

james-siteclick:
`WATER-3190`

Adds UI to delete an invoice licence within an invoice.
Also refactors confirmation pages.


Stephan-Hough (approved):
These forms are are 99% similar do we need both? 
src/internal/views/nunjucks/billing/confirm-licence.njk
src/internal/views/nunjucks/billing/confirm-invoice-licence.njk
only line 15 is different and the data could be passed from the controller.

PR #1796, opened 2021-06-08
Remove re-bill from bill run - WATER-3104

Stephan-Hough:
feat/water-3104
-- extended delete invoice batch button from a bill run for rebilling
-- invoice view updated to include cancel reissue of rebill


src/internal/modules/billing/controllers/bill-run.js
```diff
@@ -111,11 +116,20 @@ const getBillingBatchCancel = async (request, h) => billingBatchAction(request,
 
 const postBillingBatchCancel = async (request, h) => {
   const { batchId } = request.params;
+
+  try {
+    // Reset flag for rebilling on linked batches (before cancelling)
```
Stephan-Hough, line 121: This is not in the acceptance criteria for the ticket. The user might want to cancel the batch but keep the bill flagged for reissue.

src/internal/modules/billing/controllers/bill-run.js
```diff
@@ -111,11 +116,20 @@ const getBillingBatchCancel = async (request, h) => billingBatchAction(request,
 
 const postBillingBatchCancel = async (request, h) => {
   const { batchId } = request.params;
+
+  try {
```
Stephan-Hough, line 120: This is not needed here. if the whole batch is cancelled the original invoice remains as marked for re-billing

Stephan-Hough (commented):
When you cancel a re-issue you can reuse the delete an invoice method postBillingBatchDeleteInvoice but will need to pass in two extra invoice IDs as parameters (originalInvoiceId and rebillInvoiceId). You will also have the existing invoiceId that is currently sent as a parameter from the route

If you are canceling the re-issue on the reversal invoice screen you need to pass the new rebill invoice as the rebillInvoiceId and the originalInvoiceId.

If you are cancelling the re-issue from the new rebill invoice then you pass the originalInvoiceId and the reversal invoice id as the rebillInvoiceId. 

then you following logic below;

If the originalInvoiceId and rebillInvoiceId (rebill or reversal) is not null then
1. remove rebill/reissue flag in the water service for the original invoice id
2. delete the invoice using the invoiceId (this could be the reversal or the rebill)
3. delete the invoice using the rebillInvoiceId (this could be the reversal or the rebill)

orherwise 

just delete the invoice using the invoiceId

Hope this makes sense but give me shout if this doesn't make sense

src/internal/lib/connectors/services/water/BillingBatchService.js
```diff
@@ -61,8 +61,8 @@ class BillingBatchService extends ServiceClient {
     return this.serviceRequest.post(uri, options);
   }
 
-  deleteInvoiceFromBatch (batchId, invoiceId) {
-    const uri = this.joinUrl(`/billing/batches/${batchId}/invoices/${invoiceId}`);
+  deleteInvoiceFromBatch (batchId, invoiceId, originalInvoiceId, rebillInvoiceId) {
+    const uri = this.joinUrl(`/billing/batches/${batchId}/invoices/${invoiceId}/org/${originalInvoiceId}/rebill/${rebillInvoiceId}`);
```
Stephan-Hough, line 65: this delete route is also used for deleting normal invoices that are not rebills therefore you can drop the last part /org/${originalInvoiceId}/rebill/${rebillInvoiceId} and add these as query params. They are optional in the water service so therefore only tag them on as query params when needed. 

You will need to add a check here if originalInvoiceId and rebillInvoiceId is not empty then add them as query params

src/internal/modules/billing/routes/bill-run.js
```diff
@@ -198,7 +198,7 @@ const routes = {
   },
   getBillingBatchDeleteInvoice: {
     method: 'GET',
-    path: '/billing/batch/{batchId}/delete-invoice/{invoiceId}',
+    path: '/billing/batch/{batchId}/delete-invoice/{invoiceId}/org/{originalInvoiceId}/rebill/{rebillInvoiceId}',
```
Stephan-Hough, line 201: You should remove the last part of this route /org/{originalInvoiceId}/rebill/{rebillInvoiceId}
and add originalInvoiceId and rebillInvoiceId as optional query params this will allow you to use the same route to delete normal as well as rebilled invoices.

src/internal/modules/billing/routes/bill-run.js
```diff
@@ -211,7 +211,9 @@ const routes = {
       validate: {
         params: Joi.object().keys({
           batchId: Joi.string().uuid(),
-          invoiceId: Joi.string().uuid()
+          invoiceId: Joi.string().uuid(),
+          originalInvoiceId: Joi.string().uuid(),
+          rebillInvoiceId: Joi.string().uuid()
```
Stephan-Hough, line 216: `originalInvoiceId: Joi.string().uuid().optional(),
   rebillInvoiceId: Joi.string().uuid().optional()`
should be received as query parameters instead so that you can use the same route to delete normal as well as rebilling invoices.

src/internal/modules/billing/routes/bill-run.js
```diff
@@ -235,7 +237,9 @@ const routes = {
...
```
Stephan-Hough, line 242: same as above you should remove the last part of the route and add the below as query parameters 
```
originalInvoiceId: Joi.string().uuid().optional(),
rebillInvoiceId: Joi.string().uuid().optional()
```

Stephan-Hough (commented):
Some changes required on the routes and also in the UI. Not sure some of the logic is required in the nunjucks view but will discuss.

timtamimi (approved):

PR #1817, opened 2021-06-15
Add pre-handler to redirect user for completed returns

james-siteclick:


src/external/modules/returns/pre-handlers.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const { returnStatuses } = require('shared/lib/constants');
+const returnPath = require('external/lib/return-path');
+
+const assertReturnStatusIsDue = (request, h) => {
+  if (!isDueReturn(request.model)) {
```
timtamimi, line 7: It may be healthier if `isDueReturn` would just take `request` as the input.

Because.... (Go to second comment)

src/external/modules/returns/pre-handlers.js
```diff
@@ -0,0 +1,16 @@
...
+    const { path } = returnPath.getReturnPath(request.model, request);
+    return h.redirect(path).takeover();
+  }
+  return h.continue;
+};
+
+const isDueReturn = ret => ret.status === returnStatuses.due;
```
timtamimi, line 14: You can then change this function to use `lodash.get` to minimise the chance of an error occurring due to a missing property in the request object. E.g.:

```

const isDueReturn = request => {
  const returnStatus = lodash.get(request, 'model.status', an-appropriate-default-value);
  return returnStatus === returnStatuses.due;
}

timtamimi (changes requested):

timtamimi (approved):

PR #1819, opened 2021-06-15
Feat/water 3113 UI gauging station

DashCoder:


timtamimi:
Please don't forget to sort out the code smells :D

src/internal/views/nunjucks/gauging-stations/licences.njk
```diff
@@ -0,0 +1,223 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+
+    <!-- Start tag-table.html: Licences linked to this monitoring station -->
+    {% block service %}
+        <!--setting the loop number to 0-->
+        {% set loopNumber = "" %}
+
+        <!--getting the licence data-->
+        {% set tags = data.stations[data.stationID]['tags'] %}
```
james-siteclick, line 114: Given that (I think) the page renders a single station, why does the view model contain multiple stations?

I think I would expect an array called `tags` to be output to the view model here.

src/internal/modules/gauging-stations/lib/mappers.js
```diff
@@ -0,0 +1,92 @@
...
+    stationID: 0
+  };
+  const newStations = JSON.parse(JSON.stringify(data));
+  newData.stations = newStations;
+  return newData;
+};
+
+exports.mapStationsLicences = mapStationsLicences;
+exports.mapStations = mapStations;
+exports.mapTags = mapTags;
```
james-siteclick, line 92: A general comment on the mappers - as far as I can tell from the design:
https://wrls-prototype.herokuapp.com/sandbox/station?stationID=0

It looks like we are trying to group the list of tags we get back from the service by licence number for display in the table.

Therefore I wonder if it would be simpler to do something along the lines of:

```
const { groupBy } = require('lodash');

const mapStationsLicences = data => {
  const groupedByLicence = groupBy(data, row => row.licenceRef);
  return Object.values(groupedByLicence).map(mapTagGroup);
}

const mapTagGroup = data => {
// ...
````

File: src/internal/modules/gauging-stations/controller.js [hunk 3]
```diff
@@ -0,0 +1,41 @@
...
+  let newData = mappers.mapStationsLicences(data);
+  /* Format data for stations */
+  if (newData.stations) {
+    newData = mappers.mapStations(newData);
+  }
+  /* Format data for tags */
+  if (newData.stations) {
+    newData = mappers.mapTags(newData);
+  }
+  if ((newData.stations.length === 0) || (newData.stations[0].licenceRef === undefined)) {
```
james-siteclick, line 24: If we wish to have a 404 response the returning a Boom.notFound from the handler should work.

However it may be easiest to implement by returning the Boom.notFound from the pre-handler if the station is not found.  If a pre-handler returns a 404 it takes over the HTTP response.

src/internal/views/nunjucks/gauging-stations/licences.njk
```diff
@@ -0,0 +1,223 @@
...
+{% from "govuk/components/details/macro.njk" import govukDetails %}
+{% from "govuk/components/textarea/macro.njk" import govukTextarea %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% from "badge.njk" import badge %}
+
+<!--set title-->
+{% set title = [data.stations[data.stationID].riverName] + " at " + [data.stations[data.stationID].label] %}
+
+{% block pageTitle %}
```
james-siteclick, line 14: If the `pageTitle` property is output from the controller, this shouldn't be needed

src/internal/views/nunjucks/gauging-stations/licences.njk
```diff
@@ -0,0 +1,223 @@
...
+  <!-- page title -->
+  <div class="govuk-grid-row govuk-!-margin-bottom-6">
+    <div class="govuk-grid-column-full">
+      <h1 class="govuk-heading-l govuk-!-margin-bottom-1">
+        <span class="govuk-caption-l">{{[data.stations[data.stationID].catchmentName]}}</span>
+        {{title}}
+      </h1>
+      {% set metaData %}
+        {{ govukSummaryList({
+        classes: 'govuk-summary-list--no-border',
```
james-siteclick, line 28: Similar to comment below, if this page renders a single station, why not have 

`station.`

rather than

`data.stations[data.stationId]`

in the view model?

src/internal/views/nunjucks/gauging-stations/licences.njk
```diff
@@ -0,0 +1,223 @@
...
+
+            {% for i in tags %}
+              <!--setting the loop number -->
+              {% set loopNumber = loop.index0 %}
+              <tr class="govuk-table__row">
+                <td class="govuk-table__cell">
+                  <!--set the link to go to a specific licence-->
+                  {% for licence in data['licences']%}
+                    {% if licence.number == i.licenceNumber %}
+                      <a href="licence?ID={{loop.index0}}" class="govuk-link">
```
james-siteclick, line 137: This link does not look correct, I would expect it to be in the form:

`/licences/{someId}`

I think it might be preferable to define the links in the JS mapping rather than the Nunjucks as it's easier to test the logic.

Additionally the closing `A` tag is not included in the conditional.

src/internal/views/nunjucks/gauging-stations/licences.njk
```diff
@@ -0,0 +1,223 @@
...
+                  {% endfor %}
+                  {{i.licenceNumber}}</a>
+                </td>
+
+                <!--get the tags for that individual licences-->
+                {% set licenceTags = i['tagValues'] %}
+
+                <td class="govuk-table__cell ">
+                  {% set period = licenceTags[0].abstractionPeriod %}
+                  {{ period.periodStart | abstractionPeriod }} to {{ period.periodEnd | abstractionPeriod }} <br>
```
james-siteclick, line 148: As mentioned above this could be shortened to:
```
{{ tag.abstractionPeriod | abstractionPeriod }}
```
if the service was responding with the standardised shape.

src/shared/lib/pre-handlers/gaugingstations.js
```diff
@@ -0,0 +1,20 @@
...
+
+const { errorHandler } = require('./lib/error-handler');
+const GaugingStationsService = require('shared/lib/connectors/services/water/GaugingStationsService');
+/**
+ * Loads a gauging station from the water service using the Gauging Station Id
+ * which can be passed in or retrieved from the request object
+ * @param {Object} request
+ * @param {String} companyId guid, if provided
+ */
+const loadGaugingStations = async (request, h, gaugingStationId = null) => {
```
james-siteclick, line 11: If this is a pre-handler, how would the third argument here be provided?  The signature is usually just `(request, h)`

src/shared/lib/pre-handlers/gaugingstations.js
```diff
@@ -0,0 +1,20 @@
...
+  const id = gaugingStationId || request.params.gaugingStationId;
+  try {
+    const service = new GaugingStationsService(request.services.water);
+    return service.getGaugingStationLicences(id);
+  } catch (err) {
+    return errorHandler(err, `Licences not found for gaugingStationId: ${id}`);
```
james-siteclick, line 17: I think it would probably be good to check for a 404 and return a Boom.notFound here

james-siteclick (changes requested):

src/internal/views/nunjucks/gauging-stations/licences.njk
DashCoder: Commented out. Uncompleted design related to selecting a particular licence from the list.

src/internal/modules/gauging-stations/lib/mappers.js
DashCoder: Above looks like an elegant solution. At this stage I'm not sure whether one sql join will work in the long run. E.g. larger datatables may benefit from two async calls and rejoining data in UI by licenceId. I suggest merging 3113 with the other gauging stations tickets and review all together this week.

src/internal/modules/gauging-stations/controller.js
DashCoder: loadGaugingStations in pre-handler is returning a Boom.notFound now. The CI scripts is calling above with validation of gaugingstationId.

src/internal/views/nunjucks/gauging-stations/licences.njk
```diff
@@ -0,0 +1,221 @@
...
+            {{ govukButton({
+              text: "Remove a tag",
+              href: "monitoring-stations/" + gaugingStationId + "/remove-tag",
+              classes: "govuk-button--secondary"
+            }) }}
+          </div>
+
+        {% else %}
+          <p class="govuk-body">There are no licences tagged with restrictions for this monitoring station</p>
+          {{ govukButton({
```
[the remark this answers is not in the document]
timtamimi: Additionally, I think you're missing a forward slash at the beginning of the href. Should probably be /monitoring-stations/ rather than monitoring-stations/

Could I please also ask you to change `new-tag` to `tagging-licence`

src/shared/lib/connectors/services/water/GaugingStationsService.js
```diff
@@ -0,0 +1,10 @@
+const ServiceClient = require('../ServiceClient');
+const config = require('internal/config');
+class GaugingStationsService extends ServiceClient {
+  getGaugingStationLicences (gaugingStationId) {
+    const id = gaugingStationId;
```
timtamimi, line 5: Is there any need to define id here?

timtamimi: In other words, is it not the same as `const url = `${config.services.water}/gauging-stations/${gaugingStationId}/licences`;`

src/shared/view/nunjucks/filters/gauging-station-badge.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const badge = {
+  no_restriction: { status: 'success', text: 'no restriction' },
```
timtamimi, line 4: I **_think_** we call that `resume` rather than `no_restriction`.

This is because licences may have a status of `resume` or `null` but they'll never have a status that says `no_restriction`.

I might be wrong here so feel free to ignore this

src/internal/modules/gauging-stations/controller.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+  return h.view('nunjucks/gauging-stations/licences', {
+    ...request.view,
+    tableCaption: 'All licences for gaugingstation',
+    pageTitle: `${newData.stations[0].riverName} at ${newData.stations[0].label}`,
+    gaugingStationId: !gaugingStationId ? 0 : gaugingStationId,
+    catchmentName: newData.stations[0].catchmentName,
+    tags: tags,
+    station: newData.stations[0],
+    back: `/gaugingstation/${gaugingStationId}#back`
```
timtamimi, line 53: This `back` value here won't do anything if the user has JS disabled. Could you kindly change it to something robust like taking the user back to the search page or the licence page?

timtamimi (changes requested):
Please see the inline comments + let me know if you have any queries or anything :D

timtamimi (approved):

PR #1847, opened 2021-07-02
Changes following discussion with UCD

timtamimi:
Makes further changes to the tagging flow, following discussion with UCD. 

Original ticket was: https://eaflood.atlassian.net/browse/WATER-3113

This goes along with: https://github.com/DEFRA/water-abstraction-service/pull/1460


uijim (approved):
I've not run this locally but the code's looking good.

PR #1850, opened 2021-07-06
Fudges the licence condition selection mechanism to comply with UCD p…

timtamimi:
…attern

Relates to https://eaflood.atlassian.net/browse/WATER-3172

### DIFF
--- a/src/internal/modules/gauging-stations/controller.js
+++ b/src/internal/modules/gauging-stations/controller.js
@@ -152,11 +152,14 @@ const postCondition = async (request, h) => {
   }
 
   const condition = form.fields.find(field => field.name === 'condition');
+
   session.merge(request, {
     condition
   });
 
-  return helpers.redirectTo(request, h, condition.value ? `/check` : '/abstraction-period');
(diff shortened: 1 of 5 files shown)


src/internal/modules/gauging-stations/controller.js
```diff
@@ -152,11 +152,14 @@ const postCondition = async (request, h) => {
   }
 
   const condition = form.fields.find(field => field.name === 'condition');
+
   session.merge(request, {
     condition
   });
 
-  return helpers.redirectTo(request, h, condition.value ? `/check` : '/abstraction-period');
+  const conditionIsValid = condition.value !== '00000000-0000-0000-0000-000000000000';
```
uijim, line 160: Minor thing and not a blocker... do you think it's worth adding `00000000-0000-0000-0000-000000000000` to a constant somewhere so it can be reused across the various files? Don't know how things work just yet so feel free to overrule.

uijim (approved):
Approved with a minor query but not a showstopper for approval.

PR #1853, opened 2021-07-07
Update search ForLicence.spec .js

srujan-chada:
Tests updated so now Regression is covered with additional scenarios such as user searching for licenses using License Number and License Alias Name.


cypress/integration/internal/searchForLicence.spec.js
```diff
@@ -13,23 +13,41 @@ describe('search for licences as internal user', () => {
...
     // assert once the user is signed in
     cy.contains('Licences, users and returns');
 
-    // search for a license
+    // search for a license by using Name
     cy.get('#query').type('Anglian').should('be.visible');
     cy.get('.search__button').click();
     cy.contains('Licences').should('be.visible');
+    cy.get('.govuk-table__body').contains('03/28/69/0034').should('be.visible');
+    cy.get('.govuk-table__body').contains('03/28/69/0034').click();
```
timtamimi, line 30: This is **not** a test licence. Please remove this immediately. Gentle reminder that this is a public repository and we must never mix real data with the codebase.

timtamimi (changes requested):

timtamimi (approved):

PR #1856, opened 2021-07-13
Regression test/water 3219 - Sharing licence with existing and new user

srujan-chada:
This PR is for the story  - https://eaflood.atlassian.net/browse/WATER-3219 
Here we have tests for the user journeys where existing user shares their licences with an existing user and also with a new user

### DIFF
--- a/cypress/integration/external/sharing-access-with-existing-user.spec.js
+++ b/cypress/integration/external/sharing-access-with-existing-user.spec.js
@@ -24,7 +24,7 @@ describe('External user sharing license access with another external user ', () 
     cy.get('.licence-result__column > a').contains('AT/CURR/DAILY/01').should('be.visible');
     cy.get('#navbar-manage').contains('Add licences or give access').click();
     cy.get('.govuk-list').contains('Give or remove access to your licence information').click();
-    cy.get('.govuk-grid-column-two-thirds > .govuk-button').click();
+    cy.get('.govuk-button').contains('Give access').click();
     cy.fixture('users.json').then(users => {
       cy.get('input#email').type(users.externalAccessSharing);
     });
--- a/cypress/integration/external/sharing-access-with-existing-user.spec.js
+++ b/cypress/integration/external/sharing-access-with-existing-user.spec.js
@@ -31,7 +31,7 @@ describe('External user sharing license access with another external user ', () 
     cy.get('.form > .govuk-button').click();
     cy.get('.govuk-link').contains('Return to give access').click();
     cy.get('#signout').click();
-    //
+    // User will login and check the assigned licence
     describe('Login to check the license assigned ', () => {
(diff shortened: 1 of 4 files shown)


cypress/integration/external/sharing-access-with-existing-user.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      cy.get('input#email').type(users.external);
+    });
+    cy.get('input#password').type(Cypress.env('DEFAULT_PASSWORD'));
+    cy.get('.govuk-button.govuk-button--start').click();
+
+    //  view and assign the licence
+    cy.get('.licence-result__column > a').contains('AT/CURR/DAILY/01').should('be.visible');
+    cy.get('#navbar-manage').contains('Add licences or give access').click();
+    cy.get('.govuk-list').contains('Give or remove access to your licence information').click();
+    cy.get('.govuk-grid-column-two-thirds > .govuk-button').click();
```
timtamimi, line 27: Here on line 27 you are picking up on the button by its classname. Instead, I think you ought to identify it based on its text, or by URL. This is so that if the position of the link changes in the future, the test shouldn't necessarily fail.

cypress/integration/external/sharing-access-with-existing-user.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    cy.fixture('users.json').then(users => {
+      cy.get('input#email').type(users.externalAccessSharing);
+    });
+    cy.get('.form > .govuk-button').click();
+    cy.get('.govuk-link').contains('Return to give access').click();
+    cy.get('#signout').click();
+    //
```
timtamimi, line 34: What's this line doing here?

cypress/integration/external/sharing-access-with-new-user.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      cy.get('input#email').type(users.external);
+    });
+    cy.get('input#password').type(Cypress.env('DEFAULT_PASSWORD'));
+    cy.get('.govuk-button.govuk-button--start').click();
+
+    //  view and assign the licence
+    cy.get('.licence-result__column > a').contains('AT/CURR/DAILY/01').should('be.visible');
+    cy.get('#navbar-manage').contains('Add licences or give access').click();
+    cy.get('.govuk-list').contains('Give or remove access to your licence information').click();
+    cy.get('.govuk-grid-column-two-thirds > .govuk-button').click();
```
timtamimi, line 27: Same as above. Please identify URLs based on text or href, not by their CSS position or classname.

timtamimi (approved):

timtamimi (changes requested):
Please see the comments

timtamimi (approved):

PR #1861, opened 2021-07-15
added two journeys for returns status checks

harish-kancharla:


cypress/integration/external/user-returns-view.spec.js
```diff
@@ -0,0 +1,62 @@
+
+const { setUp, tearDown } = require('../../support/setup');
+// const checkErrorMessage = (message) => {
```
Stephan-Hough, line 3: Please remove the commented code

cypress/integration/external/user-returns-view.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    describe('it clicks on the returns tab link', () => {
+      cy.get('#tab_returns').click();
+    });
+    describe('sees the returns table', () => {
+      cy.get('#returns').should('be.visible');
+    });
+    describe('sees the status column', () => {
+      cy.get(':nth-child(1) > :nth-child(4) > .govuk-tag').should('be.visible').and('contain.text', 'Due');
+      cy.get(':nth-child(2) > :nth-child(4) > .govuk-tag').should('be.visible').and('contain.text', 'Overdue');
+    });
```
Stephan-Hough, line 60: I think it is better to be more explicit with what you look for in the page using the tbody tag just incase the header or somewhere else in the page another govuk-tag is added. 
Also we should check only the correct data is visible to external user by checking there are only 4 rows and checking the status for each row becuase you do not check for the completed return tag.
For this licence we know there are some void returns so perhaps doing an explicit positive check will be better or you could check if the void status tag is not found? 
i.e. 
```
 describe('see the returns table with 4 rows', () => {
      cy.get('tbody').find('tr').should('have.length', 4)
      describe('each row has the correct status tag', () => {
      cy.get('tbody').find(':nth-child(1) > :nth-child(4) > .govuk-tag').should('be.visible').and('contain.text', 'Due');
      cy.get('tbody').find(':nth-child(2) > ... 'Overdue');
      cy.get('tbody').find(':nth-child(3) > ... 'Complete');
      cy.get('tbody').find(':nth-child(4) > ... 'Overdue');
      });
    });
```

cypress/integration/internal/returns-status-view.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    describe('it clicks on the returns tab link', () => {
+      cy.get('#tab_returns').click();
+    });
+    describe('sees the returns table', () => {
+      cy.get('#returns').should('be.visible');
+    });
+    describe('sees the status column', () => {
+      cy.get(':nth-child(1) > :nth-child(4) > .govuk-tag').should('be.visible').and('contain.text', 'Due');
```
Stephan-Hough, line 44: Like the above comment it might be better to get the tbody tag then use the find method to only look for elements in that element tree just in case tags are added somewhere else on the page

Stephan-Hough (changes requested):
I have added some comments to check all the return statuses. Let me know what you think.

Stephan-Hough (approved):

PR #1875, opened 2021-07-23
Joi Validation Fix - submit external returns 

Stephan-Hough:
Bugfix/external submit returns
This fixes joi validation errors for the external returns flow and the regression pack


uijim (approved):
👍🏻

timtamimi (approved):

PR #1884, opened 2021-07-27
Bugfix/internal user regression fixes 2

uijim:
This PR contains:

* new `createSchema` function that allows setting of global Joi options. We can extend this further as we find more use cases.
* fix of regression issue around manual address entry


Stephan-Hough (approved):
Looks good and have approved this but I wonder if we should call the schema.js file something like joi-helper-functions.js or joi.helpers.js to keep it more open if we wanted to add other functions in there

uijim:
> Looks good and have approved this but I wonder if we should call the schema.js file something like joi-helper-functions.js or joi.helpers.js to keep it more open if we wanted to add other functions in there

Yep agree and changes. Thanks @Stephan-Hough

PR #1897, opened 2021-08-03
Bugfix/WATER-3288 - Charge element quantities page failing when optional billable quantity is empty

uijim:
Ticket: https://eaflood.atlassian.net/browse/WATER-3288

This PR covers 2 things:

* fixes validation issues for optional billableAnnualQuantity
* fixes issue where optional billableAnnualQuantity was not saving to session state

### DIFF
--- a/src/internal/modules/charge-information/lib/charge-elements/mappers.js
+++ b/src/internal/modules/charge-information/lib/charge-elements/mappers.js
@@ -1,5 +1,6 @@
 'use-strict';
 
+const { isEmpty } = require('lodash');
 const { SOURCES, EIUC_SOURCE_OTHER } = require('./constants');
 
 const abstraction = (formValues) => {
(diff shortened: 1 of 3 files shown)


src/internal/modules/charge-information/lib/charge-elements/mappers.js
```diff
@@ -40,9 +39,10 @@ const source = formValues => ({
 
 const quantities = formValues => {
   const { authorisedAnnualQuantity, billableAnnualQuantity } = formValues;
+
   return {
     authorisedAnnualQuantity: parseFloat(authorisedAnnualQuantity),
-    billableAnnualQuantity: isEmpty(billableAnnualQuantity) ? null : parseFloat(billableAnnualQuantity)
+    billableAnnualQuantity: !billableAnnualQuantity ? null : parseFloat(billableAnnualQuantity)
```
timtamimi, line 45: Could you revert this please? `isEmpty` is used here because this value can sometimes receive `''` or `null` or `undefined`.

`isEmpty()` allows us to conveniently check across the board, whereas `!` will only check falsy or undefined.

timtamimi (changes requested):

src/internal/modules/charge-information/lib/charge-elements/mappers.js
uijim: Done.

timtamimi:
@uijim You requested a review on this PR but as far as I can tell, the greater(0) is still there. As previously discussed: Zeros are a valid billing quantity. Unless you've already investigated and found that greater(0) allows zeros?

timtamimi (approved):

PR #1904, opened 2021-08-11
validate one/multiple checkboxes selected

DashCoder:


src/internal/modules/gauging-stations/controller.js
```diff
@@ -320,13 +321,18 @@ const getRemoveTagsConditions = async (request, h) => {
 };
 
 const postRemoveTagsLicenceSelected = async (request, h) => {
-  const form = await formHandler.handleFormRequest(request, linkageForms.removeTagsLicenceConditions);
+  const formCheckBox = await formHandler.handleFormRequest(request, linkageForms.removeTagsLicenceConditions);
 
-  if (!form.isValid) {
-    return h.postRedirectGet(form);
+  /* selectedCondition can be either string or array */
+  const removeTagsConditionsSchema = Joi.object().keys({
```
Stephan-Hough, line 327: As discussed this is duplicating the validation of the payload received by the form. Ideally we need to fix that instead of add more code here.

Stephan-Hough (changes requested):
I left a comment on the controller as discussed

Stephan-Hough (approved):

PR #1908, opened 2021-08-13
Water - 3245 - is re-billable pre-handler fixed to follow new logic

Stephan-Hough:
feat/water-3245
-- isRebillableBill prehandler fixed to match new logic in the service

### DIFF
--- a/src/internal/modules/billing-accounts/pre-handlers.js
+++ b/src/internal/modules/billing-accounts/pre-handlers.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const Boom = require('@hapi/boom');
-const { get, sortBy, isEmpty } = require('lodash');
+const { get, sortBy } = require('lodash');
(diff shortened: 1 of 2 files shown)


src/internal/modules/billing-accounts/pre-handlers.js
```diff
@@ -85,7 +85,7 @@ const isRebillableBill = bill =>
   (bill.batch.source === 'wrls') &&
   !bill.isDeMinimis &&
   (bill.netTotal !== 0) &&
-  isEmpty(bill.originalBillingInvoiceId);
+  (bill.rebillingState === 'rebill' || isEmpty(bill.rebillingState));
```
timtamimi, line 88: Line 88 could be `['rebill', null].includes(bill.rebillingState)`

Not that it makes a huge difference.

timtamimi (approved):

PR #1909, opened 2021-08-15
rebilling ui: add links to relevant bills

DashCoder:


src/internal/views/nunjucks/billing/batch-invoice.njk
```diff
@@ -72,18 +72,90 @@ invoiceDebitsTotal %}
...
+      <!-- Bills linked to reissue -->
+      {% macro displayBillingState(inv, billingBatchId) %}
+        <!-- Is "rebill" (as opposed to "rebilled") in invoices -->
+        {% if inv.rebillingStateLabel == 'rebill' and inv.originalInvoiceId != null %}
+          Reissued <span class="screen-reader-only">bill</span>
+        {% endif %}
+
+        {% if inv.rebillingStateLabel == 'rebilled' and inv.originalInvoiceId != inv.id %}
+          Reissue <span class="screen-reader-only">bill</span>
+        {% endif %}
```
Stephan-Hough, line 84: This could be changed to 
`{% if inv.rebillingStateLabel == 'rebill' or inv.rebillingStateLabel == 'rebilled' %}
          Reissued <span class="screen-reader-only">bill</span>
  {% endif %}`

src/internal/views/nunjucks/billing/batch-invoice.njk
```diff
@@ -72,18 +72,85 @@ invoiceDebitsTotal %}
...
+      {% macro displayBillingState(inv, billingBatchId) %}
+        {% if inv.rebillingStateLabel == 'rebill' or inv.rebillingStateLabel == 'rebilled'  %}
+          Reissued <span class="screen-reader-only">bill</span>
+        {% endif %}
+
+        {% if inv.rebillingStateLabel == 'original' %}
+          Original <span class="screen-reader-only">bill</span>
+        {% endif %}
+
+        {% if inv.rebillingStateLabel == 'reversal' and inv.originalInvoiceId != null %}
```
Stephan-Hough, line 85: this 
`{% if inv.rebillingStateLabel == 'reversal' and inv.originalInvoiceId != null %}`
can change to
`{% if inv.rebillingStateLabel == 'reversal' %}`
because if the label is reversal it is the reversal and the originalInvoiceId can never be null

src/internal/views/nunjucks/billing/batch-invoice.njk
```diff
@@ -72,18 +72,85 @@ invoiceDebitsTotal %}
...
+          </p>
+          {{ invoiceAddress(invoice, invoice.invoiceAccount.company) }}
         {% endset %}
         {{ govukDetails({
           summaryText: "Billing account details",
           html: html })
         }}
+
+        {% if batch.status == 'sent' %}
+          {% if (invoice.linkedInvoices.length) or (invoice.rebillingState == 'rebilled') %}
```
Stephan-Hough, line 147: I think this or `(invoice.rebillingState == 'rebilled')` can be removed from the if statement. 
If there are linked invoices then they will always relate to rebilling.

Stephan-Hough (changes requested):
just a couple of minor suggested changes

PR #1911, opened 2021-08-20
bugxif - water 3245 - rebilling

Stephan-Hough:
bugfix/water-3245
-- max rebilling date fails when rebillable bills empty
this only happens when the billing account has had
no bills generated in the service.
-- custom error added to display a warning when this happens


Stephan-Hough:
design changed no longer required

PR #1960, opened 2021-11-02
bugfix - bill run caption fix

Stephan-Hough:


timtamimi (approved):
Approved but need to check the regression pack to ensure there are no broken tests

PR #1970, opened 2021-11-10
WATER-3349: Delete a company contact

defra-contractor:
See: https://eaflood.atlassian.net/browse/WATER-3349

Please note this PR is dependant upon PRs:
- https://github.com/DEFRA/water-abstraction-service/pull/1564
- https://github.com/DEFRA/water-abstraction-tactical-crm/pull/873

### DIFF
--- a/src/internal/modules/customers/controllers.js
+++ b/src/internal/modules/customers/controllers.js
@@ -7,6 +7,7 @@ const forms = require('./forms');
 const formsHelper = require('shared/lib/forms');
 const formHandler = require('shared/lib/form-handler');
 const { hasScope } = require('../../lib/permissions');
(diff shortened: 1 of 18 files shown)


timtamimi (commented):
It would probably make sense to create a new object called `contactSources` instead of reusing `addressSources`. But this is a minor nitpick so happy to leave this to another PR/another time.

The other bigger issue (not your problem - more of a question for UCD) is that removing the contact in NALD isn't going to remove it from WRLS. This is in relation to the warning messages in `getWarningText`.

Additionally, billing contacts aren't usually shown here. They're associated to the billing account company, not the licence holder company.

timtamimi:
We agreed with Chris that the warning message on the deletion page will be changed to the following:

> Only additional contacts can be removed from this customer

PR #1978, opened 2021-11-18
WATER-3371: Recalculating charges with no change to charge versions

defra-contractor:
See: https://eaflood.atlassian.net/browse/WATER-3371

Please note that this is dependant on https://eaflood.atlassian.net/browse/WATER-3373 being complete.


Stephan-Hough (changes requested):
I think you might have accidentally renamed this file. cypress/integration/internal/remove-individual-licence-from-bill.spec.js

Stephan-Hough (approved):

PR #1981, opened 2021-11-22
Update notify-callback.spec.js

srujan-chada:
Added couple of seconds timer to make this test (notify-call-back.spec) stable


timtamimi (changes requested):
Could you re-check this please? It doesn't look quite right 😅 Many thanks.

PR #1992, opened 2021-12-13
regression test/water 3282

DashCoder:


mo-zag (approved):

srujan-chada (changes requested):
I request to add some negative scenarios to check the below scenarios - 
Both Passwords (Entered and Confirmed) don't match
Passwords does not follow the password combination like numbers and letters

Stephan-Hough (approved):

PR #1998, opened 2021-12-20
Water 3398 - SRoC charge category flow

Stephan-Hough:
Feat/water 3398


src/internal/modules/charge-information/controllers/charge-information.js
```diff
@@ -286,6 +286,28 @@ const redirectToStartOfElementFlow = (request, h) => {
...
 
+const redirectToStartOfCategoryFlow = (request, h) => {
+  const { chargeVersionWorkflowId } = request.query;
+  const { licenceId } = request.params;
+  const { draftChargeInformation: currentState } = request.pre;
+
+  // Create new element to edit in the session state
+  const id = uuid();
+  const data = currentState.chargeElements.reduce((acc, element) => {
+    element.scheme === 'alcs'
```
timtamimi, line 297: I'm not 100% sure what's happening here.

if the element has a scheme value of `alcs`, why are we doing `acc.chargePurposes.push({ ...element, scheme: 'sroc' })`?

Is this a post-SROC only thing? If so, is it worth adding a condition here to check the isSrocLive flag?

src/internal/modules/charge-information/lib/actions.js
```diff
@@ -66,9 +69,9 @@ const setAbstractionData = (request, formValues) => {
 };
 
 const getNewChargeElementData = (request, formValues) => {
-  const { defaultCharges } = request.pre;
+  const { defaultCharges, draftChargeInformation } = request.pre;
   const { step } = request.params;
-  return mappers[step] ? mappers[step](formValues, defaultCharges) : omit(formValues, 'csrf_token');
+  return mappers[step] && draftChargeInformation.scheme !== 'sroc' ? mappers[step](formValues, defaultCharges) : omit(formValues, 'csrf_token');
```
timtamimi, line 74: What's happening here? Why do we omit `csrf_token`? Could you kindly add a comment for future reference?

src/internal/modules/charge-information/lib/charge-categories/constants.js
```diff
@@ -0,0 +1,111 @@
...
+  tier2: 'tier 2'
+};
+
+const LOSS_CATEGORIES = {
+  high: 'high',
+  medium: 'medium',
+  low: 'low'
+};
+
+const WATER_AVAILABILITY = {
```
timtamimi, line 35: I mentioned in the service pull request that I'm not 100% sure this value needs to be a string as it could simply be represented by true/false.

Not something I feel strongly about - just thought I'd mention it here too in case it's of interest.

src/internal/modules/charge-information/lib/charge-categories/constants.js
```diff
@@ -0,0 +1,111 @@
...
+  available: 'available',
+  restricted: 'restricted availablity or no availability'
+};
+
+const YES_NO = {
+  // true: true,
```
timtamimi, line 41: What is this doing? the //true: true doesn't look right. Could you kindly re-check what's happening here?

src/internal/modules/charge-information/lib/charge-information-validator.js
```diff
@@ -37,8 +37,18 @@ const validate = chargeElement =>
...
+        ...element,
+        chargePurposes: element.chargePurposes.map(purpose => ({
+          ...purpose,
+          validationWarnings: validate(purpose)
+        }))
+      }
+      : element),
+    validationWarnings: chargeInformation.scheme === 'alcs'
+      ? validate(element)
+      : [] // a new validator for @SROC will have to be added here
```
timtamimi, line 51: Please add `todo` in the comment so that SonarCloud will flag it up in the future pull request

timtamimi: Info here https://rules.sonarsource.com/go/RSPEC-1135

timtamimi (changes requested):

src/internal/modules/charge-information/lib/charge-categories/constants.js
Stephan-Hough: changed to boolean and renamed property to isRestrictedSource to match the water service

src/internal/modules/charge-information/lib/actions.js
Stephan-Hough: done

src/internal/modules/charge-information/controllers/charge-element.js
```diff
@@ -16,7 +16,7 @@ const actions = require('../lib/actions');
 const getBackLink = request => {
   const { step, licenceId, elementId } = request.params;
   const { chargeVersionWorkflowId } = request.query;
-  if (request.query.returnToCheckData === 1) {
+  if (request.query.returnToCheckData) {
```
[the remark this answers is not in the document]
Stephan-Hough: When I was testing this part it failed and appeared to be a bug because the query param returnToCheckData is defined as a boolean value at the route with a default value of false. That's the reason for the change.

src/internal/modules/charge-information/controllers/charge-information.js
Stephan-Hough: Yes this is a SROC controller method. The access to this method is controlled via the SROC feature flag on the view.

src/internal/modules/charge-information/lib/charge-categories/constants.js
Stephan-Hough: changed this to the shape below to act as a double purpose mapper and feed the options on the form. 
`const YES_NO = [
  {
    label: 'yes',
    value: true
  },
  {
    label: 'no',
    value: false
  }
];`

src/internal/modules/charge-information/lib/charge-information-validator.js
Stephan-Hough: ah nice one i basically removed todo to remove the sonarcloud code smell added it back in.

PR #2006, opened 2022-01-05
deleted duplicated tests and renamed the other to best suit the test

srujan-chada:
https://eaflood.atlassian.net/browse/WATER-3358

Deleted the duplicate and renamed the test to suit the test


timtamimi (approved):
Approved, but before merging could you kindly rename the test so that the name of the test starting with billing- (Just so that it gets grouped with the other billing tests)

PR #2015, opened 2022-01-18
WATER-3393

mo-zag:
Changes to validation for charge start date.
Display latest licence version start date instead of license start date.
validate start date to allow Only if there is a licence holder within custom start date entered by user.


src/internal/modules/charge-information/routes/charge-information.js
```diff
@@ -99,7 +99,9 @@ module.exports = {
       pre: [
         { method: preHandlers.loadDraftChargeInformation, assign: 'draftChargeInformation' },
         { method: preHandlers.loadLicence, assign: 'licence' },
-        { method: preHandlers.loadIsChargeable, assign: 'isChargeable' }
+        { method: preHandlers.loadIsChargeable, assign: 'isChargeable' },
+        { method: preHandlers.loadLicenceVersions, assign: 'licenceVersions' },
```
Stephan-Hough, line 103: the licenceVersions naming here is misleading because you are returning a single object (the oldest licence version) and not an array.

src/internal/modules/charge-information/forms/start-date.js
```diff
@@ -44,13 +44,21 @@ const getDates = licence => {
...
+    minDate: isLicenceStart ? licenceDocumentsRoles.startDate : minDate.format(ISO_FORMAT),
+    minType: MIN_LICENCE_START,
+    maxDate: licence.endDate || '3000-01-01'
+  };
 };
 
 const getCommomCustomDateErrors = dates => ({
   'date.min': {
-    message: minErrors[dates.minType]
+    message: `You must enter a date after this date ${moment(dates.licenceStartDate).format('DD MMM YYYY')}`
```
Stephan-Hough, line 61: This error message is ambiguous. The user will not know what date it is I.e. licence start date or licence version start date or something else. Probably best to run this past UCD

src/internal/modules/charge-information/forms/start-date.js
```diff
@@ -44,13 +44,21 @@ const getDates = licence => {
   };
 };
 
-const minErrors = {
-  [MIN_LICENCE_START]: 'You must enter a date after the licence start date'
+const getDatesWithDocumentRoles = (licence, licenceDocumentsRoles) => {
+  const startDate = moment(licenceDocumentsRoles.startDate);
+  const minDate = startDate;
+  const isLicenceStart = startDate.isAfter(minDate);
```
Stephan-Hough, line 50: Thi block of code here doesn't make sense. If startDate = 1 Jan 2001 and minDate = startDate then startDate will never be after minDate because minDate is also 1 Jan 2001.

timtamimi:
Superceded https://github.com/DEFRA/water-abstraction-ui/pull/2026

PR #2016, opened 2022-01-19
Feat/water 3398 bugfixes round 2

Stephan-Hough:
-- bug fixes for the back button issues
-- added a temp status flag to charge elements to filter them out if data is incomplete


mo-zag (approved):
All looks good just a not if easier in future releases maybe we can look to add 'alcs' into a global const just a suggestion one for the future enhancement.

src/internal/modules/charge-information/forms/charge-element/quantities.js
```diff
@@ -8,16 +8,18 @@ const { getChargeElementData, getChargeElementActionUrl } = require('../../lib/f
 
 const getErrors = key => {
   const errors = {
-    'string.pattern.base': {
+    'number.unsafe': {
+      message: `Enter a number for the ${key} quantity using 6 decimal places or fewer, the number must be more than 0`
```
defra-contractor, line 12: Perhaps set the message as a constant and reuse for both number.unsafe and number.custom.

src/internal/modules/charge-information/pre-handlers.js
```diff
@@ -32,6 +32,22 @@ const getFilteredChangeReasons = async type => {
...
+ * Removes incomplete charge elements to avoid breaking the UI
+ * and resets the session draftCharge info
+ * @param {*} request
+ */
+const loadValidatedDraftChargeInformation = async request => {
+  const { licenceId } = request.params;
+  const chargeVersionWorkFlowId = getChargeVersionWorkflowId(request);
+  const draftChargeInformation = await loadDraftChargeInformation(request);
+  // filter out incomplete charge elements when they have used the back button
+  draftChargeInformation.chargeElements = draftChargeInformation.chargeElements.filter(element => isEmpty(element.status));
```
defra-contractor, line 45: Is element.status normally an array or object? If not, is the isEmpty function correct here?

src/internal/views/nunjucks/charge-information/macros/charge-element-table-sroc.njk
```diff
@@ -94,14 +94,14 @@ formErrorSummary %}
       categoryId: categoryId
     } | queryString %}
   {% endif %}
-  {% if chargeElementId %}
-    {{ changeLinkRoot + '/charge-element/' + chargeElementId + urlTail + '?' + queryTail }}
+  {% if chargePurposeId %}
```
defra-contractor, line 97: Is this link correct for charge purpose?

test/internal/modules/charge-information/lib/actions.js
```diff
@@ -238,6 +267,37 @@ experiment('internal/modules/charge-information/lib/actions', () => {
           }]
         });
       });
+      test('when the charge element step is purpose then set the satus to draft', () => {
```
defra-contractor, line 270: typo "satus" should be "status"

test/internal/modules/charge-information/lib/actions.js
```diff
@@ -238,6 +267,37 @@ experiment('internal/modules/charge-information/lib/actions', () => {
...
+            source: 'supported',
+            scheme: 'alcs',
+            purposePrimary: {},
+            purposeSecondary: {},
+            purposeUse: { id: 'test-purpose-id' },
+            status: 'draft'
+          }]
+        });
+      });
+      test('when the charge element step is loss then set the satus is removed', () => {
```
defra-contractor, line 287: typo satus should be status

test/internal/modules/charge-information/pre-handlers.js
```diff
@@ -531,6 +592,8 @@ experiment('internal/modules/charge-information/pre-handlers', () => {
         const [,, chargeVersion] = request.setDraftChargeInformation.lastCall.args;
         expect(chargeVersion.status).to.equal('review');
         expect(chargeVersion.invoiceAccount.invoiceAccountAddress).to.equal('test-invoice-account-address-id');
+        expect(chargeVersion.chargeElements[0].id.length === 36).to.be.true();
```
defra-contractor, line 595: Would it not be better to expect the id to be equal to a string and not just the length of the string?

defra-contractor (changes requested):

src/internal/views/nunjucks/charge-information/macros/charge-element-table-sroc.njk
Stephan-Hough: yes that is correct will go through the logic with you. not ideal but it is basically re-using the charge element flow for sroc charge purposes which is one and the same thing

test/internal/modules/charge-information/pre-handlers.js
Stephan-Hough: changed this to use Joi as Tim suggested thanks

src/internal/modules/charge-information/pre-handlers.js
Stephan-Hough: Yeah it will be either exist or not as a property of charge element. So if status is present then it will discard that item in the array.

defra-contractor (approved):

PR #2017, opened 2022-01-20
Tech-Debt: Remove and update dependencies

defra-contractor:
The following:

- Remove redundant dependencies
- Update dependencies
- Fix linting errors


src/internal/modules/abstraction-reform/lib/action-creators.js
```diff
@@ -92,7 +92,7 @@ const createEditCondition = (data, defra, id) => {
  * @param {Object} user - the current user of the application
  * @return {Object} action to edit condition
  */
-const createSetStatus = (status, notes, defra) => {
+const createSetStatus = (status, notes, user) => {
```
Stephan-Hough, line 95: You changed defra to user here but in the rest of the file it remains as defra?
Perhaps this line should change instead
`@param {Object} user - the current user of the application` to 
`@param {Object} defra - the current user of the application`

either way no big deal but probably best to be consistent in the file

src/internal/modules/abstraction-reform/lib/action-creators.js
```diff
@@ -102,7 +102,7 @@ const createSetStatus = (status, notes, defra) => {
     payload: {
       status,
       notes: notes.trim() || null,
-      user: formatUser(defra),
+      user: formatUser(user),
```
Stephan-Hough, line 105: as above

Stephan-Hough (approved):

PR #2035, opened 2022-02-04
Further modifications following discussion with PO and UCD

timtamimi:
https://eaflood.atlassian.net/browse/WATER-3393


defra-contractor (approved):
These changes look ok to me.

PR #2039, opened 2022-02-11
WATER-3458: Select multiple charge elements screen

defra-contractor:
See: https://eaflood.atlassian.net/browse/WATER-3458

### DIFF
--- a/src/internal/modules/charge-information/lib/helpers.js
+++ b/src/internal/modules/charge-information/lib/helpers.js
@@ -8,6 +8,8 @@ const services = require('../../../lib/connectors/services');
 
 const { reducer } = require('./reducer');
 const routing = require('../lib/routing');
+const { ROUTING_CONFIG } = require('../lib/charge-categories/constants');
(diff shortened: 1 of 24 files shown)


src/internal/modules/charge-information/lib/helpers.js
```diff
@@ -140,11 +155,81 @@ const isOverridingChargeVersion = async (request, draftChargeVersionStartDate) =
...
+  };
+};
+
+const restoreChargeElements = async chargeElements => {
+  // filter out incomplete charge elements when they have used the back button
+  const completeChargeElements = await Promise.all(chargeElements
+    .filter(element => !element.status)
+    .map(async element => {
+      if (element.scheme === 'sroc') {
+        element.chargeCategory = await findChargeCategory(element);
```
Stephan-Hough, line 205: I don't think you need to lookup the charge category here because the pre-handler on the getCheckAnswers route will do that.

Stephan-Hough (approved):

PR #2070, opened 2022-03-21
Bugfix/water 3552 - duplicate for the release 2.20

Stephan-Hough:


jonathangoulding (approved):
No outstanding issues. 

I feel the use of lodash is a bit excessive but something for another day.

PR #2081, opened 2022-04-01
feat: generic validation helper function

jonathangoulding:
check inline error
check summary error

group context pages and assert titles

### DIFF
--- a/cypress/integration/internal/billing-rebilling-supplementary.spec.js
+++ b/cypress/integration/internal/billing-rebilling-supplementary.spec.js
@@ -85,7 +85,7 @@ describe('rebilling supplementary bill run', () => {
         // confirm rebill
         cy.get('form > .govuk-button').click();
         // asserting the bill reissue
-        cy.get('.govuk-panel').should('contain', 'You’ve marked 7 bills for reissue');
+        cy.get('.govuk-panel').should('contain', 'You’ve marked 6 bills for reissue');
       });
       describe('user generates the supplementary bill for the marked rebilling licences', () => {
         cy.get('.govuk-grid-column-full > .govuk-button').contains('Create a supplementary bill run').should('be.visible');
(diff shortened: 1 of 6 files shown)


cypress/integration/internal/sroc-charge-version-flow.spec.js
```diff
@@ -41,89 +45,223 @@ describe('Create SRoC Charge version workflow journey', () => {
       cy.get('.govuk-button').contains('Set up a new charge').click();
     });
 
-    describe('user selects reason for new charge information', () => {
-      cy.get('.govuk-heading-l').contains('Select reason for new charge information');
-      cy.get('[type="radio"]#reason-12').click();
-      cy.get('button.govuk-button').click();
+    describe('Select reason for new charge information', () => {
```
jonathangoulding, line 48: Describe relates to the page being loaded with a reference to the page title. 

Should allow easier reading, collapsing in a text editor to make the file smaller. 

can be one line:  
validateRadioOptions('Select reason for new charge information', 'reason-12', 'Select a reason for new charge information');

cypress/support/validation.js
```diff
@@ -0,0 +1,37 @@
+const checkInlineAndSummaryErrorMessage = (expectedMessage) => {
+    cy.get('.govuk-error-summary__list').contains(expectedMessage);
```
jonathangoulding, line 2: Screenshot 2022-04-01 at 11 37 52

jonathangoulding: can generically target summary and inline error messages. assuming they are the same as per GDS.

cypress/integration/internal/sroc-charge-version-flow.spec.js
jonathangoulding: https://design-system.service.gov.uk/patterns/question-pages/

All Page titles should be unique. 

Introducing them in this way allows us to establish un expected reuse of the title and to wrap each page to be tested in a describe block with the title as a unique key.

Stephan-Hough (approved):

PR #2082, opened 2022-04-04
fix: lint issues

jonathangoulding:


src/external/modules/manage-licences/controller.js
```diff
@@ -23,7 +23,7 @@ async function getManageLicences (request, h) {
  */
 const createAccessListViewModel = licenceAccess => {
   const userRoles = licenceAccess.filter(r => r.role === 'user');
-  const mapped = userRoles.map(ur => {
+  return userRoles.map(ur => {
```
jonathangoulding, line 26: @defra-contractor Interesting change.

defra-contractor: Sonar cloud complains about unnecessary variables

PR #2090, opened 2022-04-06
Amend `No returns received` message and conditionality

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3446

Currently the two-part tariff page displays `0Ml` for reported return if no return has been received. This could be misleading so we change it to display `No returns received` instead.


StuAA78:
Closed due to error with branch and new PR opened

PR #2097, opened 2022-04-21
Bugfix/water-3632 - allow charge category volume of 0ML

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3632

During QA we found that we weren't accepting a value of `0` for charge category volume. This came down to an error in the validation -- we changed the minimum to `0` but didn't realise that the `.positive()` rule still prevented `0` from being allowed. Since we are enforcing positive values simply by having a minimum, we remove this rule. We also add an extra Cypress test to confirm that `0` is accepted.

### DIFF
--- a/cypress/integration/internal/sroc-charge-version-flow.spec.js
+++ b/cypress/integration/internal/sroc-charge-version-flow.spec.js
@@ -208,8 +208,15 @@ describe('Create SRoC Charge version workflow journey', () => {
           checkNoErrorMessage();
           cy.get('.govuk-back-link').click();
         });
-        describe('user inputs amount', () => {
+        describe('user inputs value of 0', () => {
           // We use .clear() to delete any previously accepted input
+          cy.get('#volume').clear().type('0');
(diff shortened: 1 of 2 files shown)


test/internal/modules/charge-information/forms/charge-categories/volume.js
```diff
@@ -115,10 +115,10 @@ experiment('internal/modules/charge-information/forms/charge-category/volume', (
         expect(result.error).to.exist();
       });
 
-      test('must not be zero', async () => {
+      test('must be greater than 0', async () => {
```
jonathangoulding, line 118: just a text change based on the ticket: 

Numeric values equal to or greater than 0

jonathangoulding (changes requested):

test/internal/modules/charge-information/forms/charge-categories/volume.js
defra-contractor: Fixed

PR #2099, opened 2022-04-26
bugfix/water-3633 - Allow `&*()` in charge reference description

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3633

We update the UI validation to accept characters `&*()` in the charge reference description. The description in the card only refers to `&()`; the ACs however state that `&*()` should be accepted, and we use these as our "single source of truth". This also matches the validation rules required for import so we can be assured that we are being consistent across both manual and automated methods of entry.

Note that the import validation mentioned in AC3 has been added to the WIP import branch, so will be active once that branch is merged.

### DIFF
--- a/cypress/integration/internal/sroc-charge-version-flow.spec.js
+++ b/cypress/integration/internal/sroc-charge-version-flow.spec.js
@@ -165,11 +165,19 @@ describe('Create SRoC Charge version workflow journey', () => {
           cy.get('#description').type('@ is an unsupported character');
           cy.get('form > .govuk-button').contains('Continue').click();
           checkInlineAndSummaryErrorMessage(
-            'The description must only include letters a-z, hyphens, numbers and be 180 characters or fewer.');
+            'You can only include letters, numbers, hyphens, the and symbol (&) and brackets. The description must be less than 181 characters.');
           cy.reload();
         });
+        describe('user description contains an unusual but supported character', () => {
+          cy.get('#description').type('* is an supported character');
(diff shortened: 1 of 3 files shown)


cypress/integration/internal/sroc-charge-version-flow.spec.js
```diff
@@ -165,11 +165,19 @@ describe('Create SRoC Charge version workflow journey', () => {
           cy.get('#description').type('@ is an unsupported character');
           cy.get('form > .govuk-button').contains('Continue').click();
           checkInlineAndSummaryErrorMessage(
-            'The description must only include letters a-z, hyphens, numbers and be 180 characters or fewer.');
+            'You can only include letters, numbers, hyphens, the and symbol (&) and brackets. The description must be less than 181 characters.');
```
jonathangoulding, line 168: no full stops in error messages

https://design-system.service.gov.uk/components/error-summary/

if does explicitly state that but that is the direction the ui is going to get consistency.

cypress/integration/internal/sroc-charge-version-flow.spec.js
```diff
@@ -165,11 +165,19 @@ describe('Create SRoC Charge version workflow journey', () => {
...
           cy.reload();
         });
+        describe('user description contains an unusual but supported character', () => {
+          cy.get('#description').type('* is an supported character');
```
jonathangoulding, line 172: this type assertion could cover all supported chars ? 

* & (etc) - are supported characters

jonathangoulding (approved):

PR #2102, opened 2022-05-10
water abstraction alerts scenarios added

srujan-chada:
Automation Scenarios added for 

https://eaflood.atlassian.net/browse/WATER-3263


Cruikshanks (changes requested):
The name of the file I think contains a typo @srujan-chada 

`sednig-water-abstraction-alerts-reduce-option.spec.js` I'm guessing should be ``sending-water-abstraction-alerts-reduce-option.spec.js` 🤷 

Other than that I'm happy this asserts the scenario you are trying to cover and from a look at other tests follows the same format.

Would anything more be expected of a reviewer for acceptance tests?

Cruikshanks (approved):

PR #2110, opened 2022-05-27
Amended the naming convention

srujan-chada:
Amended the naming convention so it matches the other tests.


shabbirshirazi (approved):

Cruikshanks:
Amended the naming convention so it matches the other tests.

PR #2119, opened 2022-07-01
Regression test/fixes

srujan-chada:
https://eaflood.atlassian.net/browse/WATER-3679

### DIFF
--- a/cypress/integration/internal/billing-cancel-bill-run.spec.js
+++ b/cypress/integration/internal/billing-cancel-bill-run.spec.js
@@ -74,6 +74,9 @@ describe('Cancel bill run', () => {
       cy.get('.govuk-radios__item').last().children().first().click()
       cy.get('button.govuk-button').click()
 
+      cy.get('#select-financial-year').click();
+      cy.get('button.govuk-button').click()
+
       cy.get('[role="button"]', { timeout: 1000 }).should('contain', 'Cancel bill run')
       cy.get('[role="button"]').click()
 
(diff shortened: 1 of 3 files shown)


Cruikshanks (dismissed):

cypress/integration/internal/billing-cancel-bill-run.spec.js
```diff
@@ -74,6 +74,9 @@ describe('Cancel bill run', () => {
       cy.get('.govuk-radios__item').last().children().first().click()
       cy.get('button.govuk-button').click()
 
+      cy.get('#select-financial-year').click();
```
Cruikshanks, line 77: We use the Defra standard linter now which does not like semi-colons

```suggestion
      cy.get('#select-financial-year').click()
```

cypress/integration/internal/sroc-select-multiple-charge-version.spec.js
```diff
@@ -114,6 +114,7 @@ describe('Create SRoC Charge version workflow journey', () => {
       cy.get('.govuk-heading-l').contains('Select loss category')
       cy.get('[type="radio"]#loss-2').click()
       cy.get('button.govuk-button').click()
+      cy.get('form > .govuk-button').click();
```
Cruikshanks, line 117: We use the Defra standard linter now which does not like semi-colons

```suggestion
      cy.get('form > .govuk-button').click()
```

Cruikshanks (changes requested):
Just a result of the change in the linter we now use. I've suggested the changes to make. You can accept them in the GitHub UI if that makes things easier.

Cruikshanks (approved):

PR #2120, opened 2022-07-01
Tidy up .../billing/services/transactions-csv.js

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3586

We came across transactions-csv.js as part of working on altering the bill run download to handle SROC bill runs.

To say it angered and upset the current dev team is a bit of an understatement. You _could_ argue calling a function to pull a couple of fields from an object has value (though in our opinion you'd be wrong!)

But there is no reason for making an _utter_ mess of known conventions around `private` and `public` methods. It is a [common practice](https://stackoverflow.com/a/4484449/6117745) in JavaScript to prefix private module methods with an underscore. This singles intent to other developers; these methods are only used internally by the module, and those without it are for use elsewhere in the code.

So, after another day of struggling through the code imagine our consternation when we come across this

```javascript
exports._getInvoiceData = getInvoiceData
exports._getInvoiceAccountData = getInvoiceAccountData
exports._getTransactionData = getTransactionData
exports._getTransactionAmounts = getTransactionAmounts
exports.createCSV = createCSV
exports.getCSVFileName = getCSVFileName
```

We are making `private` methods `public` but when we do so we are identifying them as `private`. Arrrrgh!!! 🤬😠

As far as we can see this was _solely_ for the purpose of testing the methods. _They are not referenced anywhere else!!!_ So, added to our convention-breaking we're also directly testing methods intended to be private and making them public purely to allow those tests to run. 🤯🤦

We have no words. Just know this change attempts to make things a little better 🥹

### DIFF
--- a/src/internal/modules/billing/services/transactions-csv.js
+++ b/src/internal/modules/billing/services/transactions-csv.js
@@ -71,14 +71,14 @@ const getDebitCreditLines = (value, isCredit, debitLabel, creditLabel) => {
   }
 }
 
-const _getInvoiceData = invoice => {
-  const { netAmount, isCredit } = invoice
-  return {
-    'Bill number': invoice.invoiceNumber,
-    'Financial year': invoice.financialYearEnding,
-    ...getDebitCreditLines(netAmount, isCredit, 'Invoice amount', 'Credit amount')
-  }
-}
+// const _getInvoiceData = invoice => {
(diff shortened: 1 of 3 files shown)


Jozzey (approved):
Looks good. Thanks for saving us from having to work with that old code anymore 🥇

PR #2144, opened 2022-07-15
Additional logging for integration test

StuAA78:
We are running into issues with uploading charge information files during integration testing. We are creating a short-lived branch which we will deploy from, which contains additional logging to help us diagnose the issue.


StuAA78:
We tracked this down to an issue with Silverline WAF blocking file uploads, so this can now be closed.

PR #2145, opened 2022-07-18
Fix Compensation Charge field in CSV export

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3586

The Compensation Charge field is not correctly populated with `Y` for compensation charges. This is due to the expected `isCompensationCharge` flag not being present. We therefore replicate the logic of this flag in our CSV exporter.


Jozzey (approved):
👍🏼

PR #2153, opened 2022-07-20
Postal address issue - water abstraction alerts

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3717

When creating a water abstraction alert, the alert will be sent by letter to the licence holder if no water abstraction alert contact has been added and opted in to receive email alerts.

We update the `confirm-sending-alerts` template to display the full set of address lines that we expect to receive.

Note that we also display `address_line_7` even though elsewhere we see a comment that there can only be 6 address lines plus postcode. We do this because the notifications report is set to display an `address_line_7` so even though we don't expect to receive it, we use it anyway just in case.


Jozzey (approved):
👍🏼

PR #2163, opened 2022-07-28
Skip public water question page if licence holder isn't water undertaker

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3702

When creating SRoC charge information, we want to skip the 'is abstraction for the supply of public water' question page if the licence holder is not a water undertaker. If we do skip this page then we ensure that `isSupplyPublicWater` is defaulted to `false`, by checking if it's been set by the time we get to the adjustments page at the end of the flow.


Jozzey (approved):
Excellent work 🥇

PR #2196, opened 2022-09-02
Rename test files from filename.js to filename.test.js

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/27

VS Code puts a nice distinctive icon next to test files which are named in the format filename.test.js. This makes it easy to identify which files can be ignored when searching through the code for things.

Unfortunately, our test files are named simply filename.js, so we don't get this icon when we search and can't easily identify the test files we can ignore.

We should therefore consider renaming all the test files in the format filename.test.js.


Cruikshanks (approved):

Jozzey:
SonarCloud tests are failing as the genius that wrote some of the unit tests thought the best way to get them to pass was not to add any assertions. An issue will be raised to fix these dodgy tests in another PR

PR #2201, opened 2022-09-09
Test fixes

shabbirshirazi:
Fixing annual billing and select multiple charge version


Cruikshanks (changes requested):
It's currently failing the linting check in the build CI. However, it can all be resolved by just deleting these lines.

It's a principle that we don't comment out code. So, if these lines are truly not needed we'd ask for them to just be deleted.

Cruikshanks (approved):

PR #2213, opened 2022-10-05
Fix acceptance-tests proxy

Cruikshanks:
We thought we were blocked from being able to run our acceptance tests against our AWS environments due to the limitation that they depend on making calls to the [water-abstraction-service](https://github.com/DEFRA/water-abstraction). This is because unlike locally, only the UI is exposed in our AWS environments. A pass-through or 'proxy' would be needed to allow calls to be sent via the UI to the service API.

It seems the previous team came to the same conclusion because we found `/acceptance-tests/{tail*}` in the routes for both the internal and external app!

However, even though it's listed in both it only works in the external app instance. This change updates the implementation so that it's only accessible from the internal UI. This means we can simplify things and move the functionality into a module. We also do a bit of tidying up and add _some_ tests 😁


Cruikshanks:
I'm not hitting the test coverage mark for SonarCloud. But I think I'm being penalised because I touched some of the existing code that is missing any coverage.

Hence, I've put it out for review. But do shout if you think I should be trying a little harder 😁

Jozzey (approved):
🥇

PR #2225, opened 2022-10-25
Add new /health/info endpoint

Cruikshanks:
https://github.com/DEFRA/water-abstraction-team/issues/54

This supports the new service-status page we're building in the [water-abstraction-system repo](https://github.com/DEFRA/water-abstraction-system).

We want to be able to show both the version and commit hash for each service on that page. The existing `/status` page does include the `package.json` version. But that endpoint is also used for the AWS load balancer health check so is heavily hit. We don't want to make it do more work getting the commit hash.

Instead, we add a new `/health/` path and a new `/info` endpoint to return both the version and commit hash.

### DIFF
--- a/src/external/modules/routes.js
+++ b/src/external/modules/routes.js
@@ -1,3 +1,6 @@
+'use strict'
+
+// Internal only routes
 const addLicencesRoutes = require('./add-licences/routes')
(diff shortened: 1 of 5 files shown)


Jozzey (approved):
👍🏼

PR #2252, opened 2022-12-09
Remove hapi good - add hapi-pino

Cruikshanks:
https://github.com/DEFRA/water-abstraction-team/issues/66

[@hapi/good](https://www.npmjs.com/package/@hapi/good) is in all our app repos and is used to report stats on the running apps. The problem is it generates a horrendous amount of noise in our logs, and as of Dec 31 2020 was deprecated 😩 .

For those perfectly 'good' reasons, we need to strip it out of all repos where it is used.

The problem is it is also responsible for logging all requests and responses. Our preferred logger for this is [hapi-pino](https://github.com/pinojs/hapi-pino). It's still supported, fast, and outputs in JSON format. It also allows us to control what exactly gets logged, for example, we can suppress logging calls to `/status` which pollutes our logs in **production**.

So, this change replaces **hapi/good** with **hapi-pino**.


Cruikshanks:
I know this has failed the SonarCloud check. The reason is test coverage and the fact the plugin code has nothing. We've taken this code from [water-abstraction-system](https://github.com/DEFRA/water-asbtraction-system), which in turn took the code from [sroc-charging-module-api](https://github.com/DEFRA/sroc-charging-module-api).

So, we know it's good and that once in never gets changed. I'm comfortable letting this through but happy to discuss if anyone has any issues.

PR #2265, opened 2023-01-06
Allow charge category to be dynamic in fixtures

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3834

When working on the various scenarios needed to calculate the abstraction period for a charge version in a given billing period we realised there is one we cannot cover.

A `charge_version` can be linked to multiple `charge_elements` which can be linked to different `billing_charge_categories`. But the current fixture logic does not allow you to pick which `billing_charge_category` to link a `charge_element` to. It's hardcoded into the logic.

This change makes it possible to select a `billing_charge_category` dynamically in your fixture.


Cruikshanks:
Wrong repo! 🤦

PR #2269, opened 2023-01-17
Trigger SROC supplementary bill run in parallel

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3828

Prior to this change, we were triggering the SROC supplementary bill run when the PRESROC reported back that it was finished. The mechanism for doing this is a spinner page that polls the bill run status. When it gets out of `processing`, we're go for SROC!

But we've quickly found relying on this existing feature is flaky at best and introduces a whole bunch of other issues. Also, it doesn't work with the view bill runs screen. There is nothing to stop a user from quitting the spinner page and coming back to view bill runs. They'll see PRESROC there, but not the SROC. The existing functionality treats creation and processing separately so we should do the same for SROC.

So, this change updates the UI to trigger the creation of the SROC bill run at the same time it triggers the creation of the PRESROC. Processing the SROC bill run will still happen after the PRESROC one has been completed but those changes will happen in [water-abstraction-service](https://github.com/DEFRA/water-abstraction-service) and [water-abstraction-system](http://github.com/DEFRA/water-abstraction-system).

### DIFF
--- a/src/internal/modules/billing/controllers/create-bill-run.js
+++ b/src/internal/modules/billing/controllers/create-bill-run.js
@@ -12,6 +12,9 @@ const routing = require('../lib/routing')
 const sessionForms = require('shared/lib/session-forms')
 const { getBatchFinancialYearEnding } = require('../lib/batch-financial-year')
 
+const { logger } = require('../../../logger')
+const config = require('../../../config')
+
 /**
  * Step 1a of create billing batch flow - display form to select type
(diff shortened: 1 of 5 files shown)


Cruikshanks:
Just testing the [GitHub custom autolinks feature](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-autolinks-to-reference-external-resources)

This PR is linked to WATER-3828

PR #2270, opened 2023-01-18
Label empty bill runs as `EMPTY`

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3874

When you run a bill run for a region and the service finds no applicable licences that need billing it directs you to an error page. The page tells you the reason is that the bill run is empty, so this fact is in the data somewhere!

What we and the users would ideally like to see is the bill run labelled as `EMPTY` in the view bill runs screen. As we'll have 2 distinct supplementary bill runs being generated each time it's important we can start distinguishing between those that have truly errored, and those where there was nothing to do.

<details>
<summary>Screenshot</summary>

Screenshot 2023-01-19 at 09 52 42

</details>


Cruikshanks:
Was originally linked to WATER-3828 but we ended up creating a distinct ticket to which we have now linked the PR.

Jozzey (approved):

PR #2272, opened 2023-01-23
Add charge scheme annotation to bill runs page

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3664

With the introduction of SRoC, when running a supplementary bill run, it's possible that the bill run could pick up pre-SRoC and SRoC transactions. SOP has a constraint where they can’t deal with pre and post SRoC charges on the same bill run. Therefore supplementary billing will be split across two bill runs if there are a mixture of pre and post SRoC transactions.

To help B&D distinguish between the two charging schemes, we will add an annotation to pre-SRoC bill runs on the bill runs page - the annotation will be 'old charge scheme'.


StuAA78:
Example of UI change:

image

Cruikshanks (approved):

PR #2274, opened 2023-01-25
Add new error page for errored bill runs

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3882

When working on WATER-3833 it was tested what would happen if an error occurred.

If this happens in the legacy code the `billing_batch` has its `status` changed to error. In the UI it still gets listed with an **ERROR** label. In our new SROC supplementary bill run code though we weren't creating the record.

To ensure parity between the legacy and new code, we did some testing of what the legacy code actually does. What we found was _if_ it manages to throw and handle an error, it does set the `status` to 'error' and in some cases, it also sets an `error_code` against the bill run. But though the team thought there was an error page to display this information what actually happens is the service forwards the errored bill runs to `/billing/batch/${id}/processing`, which then 'helpfully' throws an HTTP `500`

```javascript
  // src/internal/modules/billing/controllers/bill-run.js - getBillingBatchProcessing()
  // ...

  // Render error page if batch has errored
  if (batch.status === 'error') {
    return Boom.badImplementation('Billing batch error')
  }

  // ...
```

The error page the comment is referring to is the generic 'Something's gone wrong page'. But the user will never see this because it also returns an HTTP `500` code, which our WAF blocks for good reasons (never send a 500 to the user!)

Why is this an issue now? We're implementing a process where the PRESROC and SROC bill runs are created in parallel, but processed in sequence. If an error happens during the processing of the SROC bill run that's fine because it is at the forefront. But if an error happens during creation, the PRESROC is at the forefront. We can't redirect to an error page for something happening in the background. All the user will see is the SROC bill run listed as **ERROR** with no other details.

To improve their experience, avoid needless ☎️ to the team, and stop users from seeing the Firewall error page (at least in this context) we're adding an error page, that errored bill runs can link through to.

<details>
<summary>Example screen</summary>

Screenshot 2023-01-26 at 09 40 35

</details>

### DIFF
--- a/src/internal/modules/billing/controllers/bill-run.js
+++ b/src/internal/modules/billing/controllers/bill-run.js
@@ -3,7 +3,6 @@
 const { titleCase } = require('shared/lib/string-formatter')
 const { pluralize } = require('shared/lib/pluralize')
 const moment = require('moment')
-const Boom = require('@hapi/boom')
 const { get } = require('lodash')
 
 const { cancelOrConfirmBatchForm } = require('../forms/cancel-or-confirm-batch')
--- a/src/internal/modules/billing/controllers/bill-run.js
+++ b/src/internal/modules/billing/controllers/bill-run.js
@@ -300,26 +300,22 @@ const postDeleteAllBillingData = async (request, h) => {
   return h.redirect(BATCH_LIST_ROUTE)
 }
 
-exports.getBillingBatchList = getBillingBatchList
-exports.getBillingBatchSummary = getBillingBatchSummary
-exports.getBillingBatchInvoice = getBillingBatchInvoice
-
-exports.getBillingBatchCancel = getBillingBatchCancel
-exports.postBillingBatchCancel = postBillingBatchCancel
-
-exports.getBillingBatchConfirm = getBillingBatchConfirm
-exports.postBillingBatchConfirm = postBillingBatchConfirm
-exports.getBillingBatchConfirmSuccess = getBillingBatchConfirmSuccess
-
-exports.getBillingBatchDeleteInvoice = getBillingBatchDeleteInvoice
-exports.postBillingBatchDeleteInvoice = postBillingBatchDeleteInvoice
-
-exports.getTransactionsCSV = getTransactionsCSV
-
-exports.getBillingBatchProcessing = getBillingBatchProcessing
-exports.getBillingBatchEmpty = getBillingBatchEmpty
-exports.getBillingBatchError = getBillingBatchError
-
-exports.postDeleteAllBillingData = postDeleteAllBillingData
-exports.getBillingBatchStatusToCancel = getBillingBatchStatusToCancel
-exports.postBillingBatchStatusToCancel = postBillingBatchStatusToCancel
+module.exports = {
+  getBillingBatchList,
+  getBillingBatchSummary,
+  getBillingBatchInvoice,
+  getBillingBatchCancel,
+  postBillingBatchCancel,
+  getBillingBatchConfirm,
+  postBillingBatchConfirm,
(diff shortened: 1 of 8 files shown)


src/internal/views/nunjucks/billing/batch-error.njk
```diff
@@ -0,0 +1,19 @@
...
+{% block content %}
+
+  {{ govukErrorSummary({
+    titleText: "There is a problem with this bill run",
+    errorList: errorList
+  }) }}
+
+  {{ batchHeader(pageTitle, batch) }}
+
+  <p>
```
StuAA78, line 15: Would we normally add a class to this? https://design-system.service.gov.uk/styles/typography/#paragraphs
```suggestion
  <p class="govuk-body">
```

src/internal/modules/billing/controllers/bill-run.js
```diff
@@ -257,6 +252,49 @@ const getBillingBatchEmpty = async (request, h) => {
...
+  ]
+
+  const error = errors.find((error) => {
+    if (error.code === errorCode) {
+      return true
+    }
+    return false
+  })
+
+  return [{ text: error.text }]
```
StuAA78, line 295: Are we confident that if there's an error code, it will always be one that's in the list of `errors`? Or do we want to add a little additional checking to return something like `[{ text: 'An unknown error code was assigned. We have no further information at this time.' }]`

StuAA78 (changes requested):

src/internal/modules/billing/controllers/bill-run.js
Cruikshanks: I thought about that, and TBH I want it to explode should this happen. Else I fear we'd run the risk of not exposing the fact we'd missed this should we add a new error code.

Regards the list of codes, they are all defined in the [water-abstraction-service](https://github.com/DEFRA/water-abstraction-service/blob/main/src/lib/models/batch.js#L33)

src/internal/modules/billing/controllers/bill-run.js
```diff
@@ -283,6 +321,7 @@ exports.getTransactionsCSV = getTransactionsCSV
 
 exports.getBillingBatchProcessing = getBillingBatchProcessing
 exports.getBillingBatchEmpty = getBillingBatchEmpty
+exports.getBillingBatchError = getBillingBatchError
```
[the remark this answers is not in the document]
Cruikshanks: Done!

StuAA78 (approved):

src/internal/modules/billing/controllers/bill-run.js
```diff
@@ -258,32 +253,69 @@ const getBillingBatchEmpty = async (request, h) => {
...
+    { code: 50, text: 'Error when creating the Charging Module bill run.' },
+    { code: 60, text: 'Error when deleting an invoice.' },
+    { code: 70, text: 'Error when processing two-part tariff.' },
+    { code: 80, text: 'Error when getting the Charging Module bill run summary.' },
+    { code: 90, text: 'Error when re-billing a bill run.' }
+  ]
+
+  const error = errors.find((error) => {
+    return error.code === errorCode
+  })
```
StuAA78, line 290: Just spotted your comment from last week on coding conventions and realised this is now a truthiness check we could condense to one line!
```suggestion
  const error = errors.find((error) => error.code === errorCode)
```

Cruikshanks: In my head, the exception related specifically to where you are just asserting a value is 'truthy'. i.e. it's not `null` or `undefined`.

I would still use _block_ body for the check we're doing here because I consider it an equality check rather than truth.

So, I will take your approval and run with it!

PR #2187, opened 2022-08-23
Remove unneeded dependencies

StuAA78:
https://github.com/DEFRA/water-abstraction-team/issues/26

Our `package.json` file has a number of dependencies hanging around which are no longer used. We identify them using `npx depcheck` and uninstall them accordingly.

PR #2299, opened 2023-02-24
Remove unneeded packages

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/26

We have some dependencies which aren't needed at all -- for example, they may be things that were originally in use but were since dropped, and have never been removed from `package.json`. This is slightly different to the issue of [packages which are in use but are deprecated](https://github.com/DEFRA/water-abstraction-team/issues/22) so keep the two separate.


Jozzey:
This PR replaces the stale PR at https://github.com/DEFRA/water-abstraction-ui/pull/2187

Cruikshanks (approved):

PR #2300, opened 2023-02-24
Fix licence holder display bug in check charge info

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3905

The business has raised an issue with the 'Check charge information' view. The scenario they outlined was

- Licence **AN/030/0008/006** was originally issued to _Cemex Materials Ltd_ with effect from 14/03/2022
- It was then transferred to _Breedon Trading Limited_ with effect from 27/06/2022
- Within the period 14/03/2022 - 26/06/2022 they would expect to see _Cemex Materials Ltd_ as the licence holder
- From 27/06/2022 we would expect to see _Breedon Trading Limited_ as the licence holder

This reminded someone that we'd had a similar issue previously. They found WATER-3348 which is linked to [Bugfix/water 3348](https://github.com/DEFRA/water-abstraction-ui/pull/1943). It seems that fixed the bug for the charge info view page (`/licences/{licenceId}/charge-in
formation/{chargeVersionId}/view`). But though the controller for `/licences/{licenceId}/charge-information/{chargeVersionWorkflowId}/review` is directly below it, it didn't get fixed at the same time. 🙁

Whilst digging into this we also found the same issue with the `/licences/{licenceId}/charge-information/check` screen.

So, this fix applies what was done to `getViewChargeInformation()` to `getReviewChargeInformation()` and `getCheckData()` in `src/internal/modules/charge-information/controllers/charge-information.js`.

<details>
<summary>Before fix (review)</summary>

review_charge_info_error

</details>

<details>
<summary>After fix (review)</summary>

review_charge_info_fix

</details>

### DIFF
--- a/src/internal/modules/charge-information/controllers/charge-information.js
+++ b/src/internal/modules/charge-information/controllers/charge-information.js
@@ -211,9 +211,6 @@ const getCheckData = async (request, h) => {
   const { licenceId } = request.params
   const isApprover = hasScope(request, chargeVersionWorkflowReviewer)
 
-  // const { data: documentRoles } = await services.crm.documentRoles.getDocumentRolesByDocumentRef(licence.licenceNumber)
-  // const licenceHolder = documentRoles.find(role => role.roleName === 'licenceHolder')
-
   const { data: documentRoles } = await services.crm.documentRoles.getFullHistoryOfDocumentRolesByDocumentRef(licence.licenceNumber)
   const licenceHolder = documentRoles.find(role => role.roleName === 'licenceHolder' &&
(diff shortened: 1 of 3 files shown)


Cruikshanks:
Have made an executive decision to ignore SonarCloud for now. Ideally, we'd try and come up with some unit tests that reflect the scenario and confirm our fix (instead of manual testing).

But there is nothing currently in place even for the old behaviour. We can get the fix done quickly. But adding proper unit tests is beyond the scope of this change, and the time we have for this at the moment.

Will just add it as another reason why we need to move away from this legacy code as quickly as possible!

Jozzey (approved):

PR #2303, opened 2023-03-03
Fix licence holder display bug 4 date of transfer

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3905

We had a first stab at solving how the licence holder is displayed in the charge version review and check screens in [Fix licence holder display bug in check charge info](https://github.com/DEFRA/water-abstraction-ui/pull/2300).

That basically involved applying a fix implemented by the previous team to the charge version view screen to these screens. But our meticulous QA spotted that if you created a new charge version that starts on the day of an existing transfer you don't get the wrong holder, you get nothing!

This change fixes the issue for that edge case.


Jozzey (approved):
👍🏼

PR #2324, opened 2023-03-31
Download of Bulk Returns Data Fails with Rejected URL

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3954

When trying to download the data for Bulk returns, you receive the Rejected URL error rather than the save file page.

This turns out to be because the zip file is generated as a stream in `objectMode` but hapi does not allow `objectMode` streams to be returned. We therefore need to wrap the zip file in a Readable stream to keep hapi happy.

Ideally, the zip file would not be generated as a stream in `objectMode` but we couldn't see anything in the Archiver package docs to indicate this is possible!

More info on hapi and `objectMode` streams can be found here: https://github.com/hapijs/hapi/issues/3733

### DIFF
--- a/test/external/modules/returns/lib/csv-template.test.js
+++ b/test/external/modules/returns/lib/csv-template.test.js
@@ -15,6 +15,12 @@ experiment('csv templates', () => {
   let archive
 
   beforeEach(async () => {
+    // We create an instance of `archive` so it can be piped through a passthrough stream, which would be awkward to do
+    // if we simply created a mock object.
+    archive = archiver('zip')
+    archive.append = sandbox.stub()
+    archive.finalize = sandbox.stub()
+
     sandbox.stub(logger, 'warn')
   })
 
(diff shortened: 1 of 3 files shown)


StuAA78 (dismissed):

test/external/modules/returns/lib/csv-template.test.js
```diff
@@ -372,6 +381,13 @@ experiment('csv templates', () => {
   })
 
   experiment('buildZip', () => {
+    beforeEach(async () => {
+      // Creating an actual 'archive' instance as we weren't able to wrap it as a readable stream
```
StuAA78, line 385:
```suggestion
      // We create an instance of `archive` so the function can wrap it as a readable stream, which would be awkward to
      // do if we simply created a mock object.
```

StuAA78 (approved):

PR #2327, opened 2023-04-05
Update include licence in supp. billing messages

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3948

In [Highlight flagged for SROC supplementary billing](https://github.com/DEFRA/water-abstraction-ui/pull/2322) we made a change to the notice shown to users when a licence is flagged for supplementary billing. The first iteration of the wording was what us developers had put together based on the existing messages and changes elsewhere in the service.

Our UAT team have now had a chance to look at them and asked for some tweaks.

Pre-sroc (old charge scheme) needs to be

> This licence has been marked for the next supplementary bill run for the old charge scheme.

When a licence is flagged for both the old and new schemes

> This licence has been marked for the next supplementary bill runs for the current and old charge schemes.

### DIFF
--- a/src/internal/modules/view-licences/controller.js
+++ b/src/internal/modules/view-licences/controller.js
@@ -142,11 +142,11 @@ const _includeInSupplementaryBillingMessage = (licence) => {
 
   let message = null
   if (includeInPresroc && includeInSroc) {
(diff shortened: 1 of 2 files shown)


Jozzey (approved):
👍🏼

PR #2302, opened 2023-02-28
Swap out or remove Deprecated packages

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/22

We are finding as we go through the code base that we have a number of deprecated packages.

They are no longer maintained so we need to get them out. If we don't, we expose ourselves to unfixed errors, vulnerabilities, and issues when we come to upgrade other packages.

PR #2334, opened 2023-04-24
Update package `got` & bump minor & patch versions in `package.json`

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/78

After working on the issue [Swap out or remove Deprecated packages](https://github.com/DEFRA/water-abstraction-team/issues/22) for the [water-abstraction-ui](https://github.com/DEFRA/water-abstraction-ui) in this [PR #2302](https://github.com/DEFRA/water-abstraction-ui/pull/2302) we found that in order to update/replace/remove some of the packages a significant amount of work was needed. These are therefore going to be done individually in separate PRs.

From V12 `got` is a pure ESM package so it cannot be `require()`'d from CommonJS. The following method has therefore been used to import the package `const { got } = await import('got')` to make it work with CommonJS.

I'll also get the minor and patch versions updated in the `package.json`. Since these packages are already bumped by `Dependabot` in the package-lock.json it's not really updating anything. More of a tidy up really.


Cruikshanks (approved):
Do not forget this change! 📌

We've struggled with stubbing **got** in previous projects and so have relied on web-mock. By Jove, you appear to have come up with the solution!

PR #2349, opened 2023-05-10
Enable cancel btn for empty & errored bill runs

Cruikshanks:
We've been hitting the supplementary bill run process _hard_ over the last few months as we have tested and investigated the myriad scenarios we need to cover. During this we've noted we get many more `EMPTY` bill runs because of the need to kick off both PRESROC and SROC. It has also always puzzled us why you can't get rid of `ERRORED` bill runs.

So, this quick tweak adds the **Cancel** button to the empty and errored bill run pages to allow users to get rid of them from the listings. It might also help the performance of the page not to include these bill runs in the results.

<details>
<summary>Screenshot of bill runs page</summary>

Screenshot 2023-05-10 at 09 34 25

</detail>


Jozzey (approved):
👍🏼

PR #2350, opened 2023-05-10
Increase bills runs per page

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4004

Currently, the View a Bill Run page at `/billing/batch/list` only displays a maximum of 10 records per page. The number of records displayed needs to be increased to 20.


.env.example
```diff
@@ -33,6 +33,9 @@ REDIS_PASSWORD=
 ALLOW_CHARGE_VERSION_UPLOADS=
 TRIGGER_SROC_SUPPLEMENTARY=
 
+# Defaults to 20 if not set
+RECORDS_TO_DISPLAY_PER_PAGE=
```
Cruikshanks, line 37: We display different types of results across the service but this only applies to one screen. Therefore, could we be more specific about where this env var is used?

```suggestion
BILL_RUNS_TO_DISPLAY_PER_PAGE=
```

This will have a knock-on to the config file etc but rather than fill the review will noise I'll just highlight this here.

Cruikshanks (changes requested):
Just a name change to the config value which I think will help give a clue as to where this is used/what it is for.

.env.example
Jozzey: I shall sort that out

Cruikshanks (approved):

PR #2451, opened 2023-10-10
Generate unique job ID for notifications

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4110

As part of our effort to prevent unintentional duplication of 'hands-off flow level warnings', this PR focuses on adding a unique identifier during the process. Within our notification system, there exist four distinct types of notifications, and one of these involves the 'hands-off flow warning letters'. To avoid the scenario where multiple identical requests are processed resulting in the dispatch of multiple letters or emails, the work presented in this PR aims to generate a unique job ID during the preview phase of the notifications flow. This distinctive job ID will then be sent to the "sendNotifications" job. Our system utilizes BullMQ for managing queues, workers, and jobs. BullMQ is configured in such a way that if a job with a unique ID is added to the queue, any subsequent job added with the same ID will be disregarded to prevent duplicate execution. By employing this unique ID as the job ID, we ensure that the job is not executed more than once, effectively addressing the issue of duplicate job processing.


src/internal/lib/connectors/services/water/NotificationsApiClient.js
```diff
@@ -11,7 +11,7 @@ class NotificationsApiClient extends SharedNotificationsApiClient {
    * @param {String} sender - email address of sender.  If not supplied, reverts to preview mode
    * @return {Promise} resolves with an array of contacts, each with licence numbers and rendered templates attached
    */
-  sendNotification (taskConfigId, licenceNumbers, params = {}, sender = null) {
+  sendNotification (taskConfigId, licenceNumbers, uniqueJobId, params = {}, sender = null) {
```
Cruikshanks, line 14: Someone went to the effort of adding JSDoc comments for the existing params. So, I don't think it would hurt to update them with this new one.

Cruikshanks (changes requested):
Only a minor thing to have the documentation updated as well.

I'd also suggest double checking what ChatGPT comes out with for your PR comments 😉 😂  

> This ***distinctive*** job ID will then be sent [..]

Cruikshanks (approved):

PR #2461, opened 2023-10-26
Link to new bill page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4155

We are working on implementing a new version of the bill page. The current one tries to display all transaction details for all licences on a single page when the bill is selected.

All that work is being done in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system) as part of our ongoing efforts to migrate away from the legacy code.

This means we need to update the UI to point to the system app and the endpoint we are building there when a user follows a view bill link.

We know there are 3 places we link to a bill; from the bill run summary, when looking at the bills for a selected billing account and from the licence summary 'Bills' tab. Because the work is not complete we'll put the change behind a feature flag to allow us to ship other changes if we need to.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,6 +50,7 @@ LAZY_REDIS=true
 # Feature toggles
 ENABLE_DELETE_ALL_BILLING_DATA_FEATURE=
 SHOW_VERIFICATION_CODE_FEATURE=
+USE_NEW_BILL_VIEW=
 
 # Set log level for app. Default is 'info'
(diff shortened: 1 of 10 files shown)


Cruikshanks:
Found a link I missed 🤦. Taking the PR back to sort.

Cruikshanks:
All sorted 😁

Beckyrose200 (approved):
Looks good from what I can tell :)

PR #2463, opened 2023-11-01
Add cookie to http requests to system

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4085

When we [enabled authentication by default](https://github.com/DEFRA/water-abstraction-system/pull/466) in **water-abstraction-system** we just had in mind requests proxied through or redirected from **water-abstraction-ui**.

We overlooked requests the UI was generating itself using `ServiceClient` from [water-abstraction-helpers](https://github.com/DEFRA/water-abstraction-helpers).

Doh! 🤦

So, we've inadvertently broken SROC supplementary billing and changing a billing account address because the requests we're sending to **water-abstraction-system** are being redirected to the `/signin` page.

The 'quick' fix would have been to remove auth off those routes but that would expose a vulnerability if we didn't then try to block them in `src/internal/modules/system-proxy`.

What **water0abstraction-system** needs is the cookie the UI has set once someone has been authenticated. It gets passed automatically when requests are proxied or redirected.

So, this change grabs that cookie of the [Hapi request](https://hapi.dev/api/?v=21.3.2#request) and passes it through to the HTTP request generated by **water-abstraction-helpers**. SROC supplementary billing and changing a billing account address work again!

### DIFF
--- a/src/internal/lib/connectors/services/system/BillRunsApiClient.js
+++ b/src/internal/lib/connectors/services/system/BillRunsApiClient.js
@@ -10,13 +10,17 @@ class BillRunsApiClient extends ServiceClient {
    * @param {String} scheme The scheme to create the bill run for (ie. sroc etc.)
    * @param {String} region The GUID of the region to create the bill run for
    * @param {String} user The email address of the user requesting bill run creation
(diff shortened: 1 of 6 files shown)


Jozzey (approved):
👍🏼

PR #2465, opened 2023-11-06
Link SROC 2PT bill runs in UI to new SROC review page

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4205

This PR is for changing the links on the billing/batch/list page. If a bill run is 2pt, sroc and in review then the link for the bill run should send the request to our new bill run page in the system repo.


src/internal/modules/billing/lib/routing.js
```diff
@@ -23,6 +23,13 @@ const getBillingBatchRoute = (batch, opts = {}) => {
     .set('error', `/billing/batch/${id}/error`)
     .set('queued', `/billing/batch/${id}/processing?back=${opts.isBackEnabled ? 1 : 0}`)
 
+  const routeMapTptSroc = new Map()
+    .set('review', `/system/bill-runs/${id}/review`)
+
+  if (scheme === 'sroc' && type === 'two_part_tariff' && batch.status === 'review') {
+    return routeMapTptSroc.get(batch.status)
+  }
+
```
Cruikshanks, line 32: Appreciate we can't include our new link in the existing`Map` else there would be 2 entries labelled `review`.

But as this forces us to handle this scenario outside the existing `Map` there is no need for us to create one, especially when we just have the one entry. So, I'd suggest amending the block as such

```suggestion
  if (scheme === 'sroc' && type === 'two_part_tariff' && batch.status === 'review') {
    return `/system/bill-runs/${id}/review`
  }

```

The other thing I'd suggest doing is moving this above the existing logic. In effect, we treat it as a guard clause. There is no point generating the `Map` if we never use it because we're dealing with a 2PT sroc bill run with a status of `review`.

Cruikshanks (changes requested):
I think we can simplify things and avoid unnecessary processing if we think about where we do our new check.

Cruikshanks (approved):

PR #2466, opened 2023-11-06
Update Bill Run Trigger pages for SRoC 2PT

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4197

We had to do the same thing for SROC supplementary billing. So, the existing legacy code has already been updated to send a create bill run request to `water-abstraction-system` at these points. However, it ignores anything that isn’t a supplementary. It needs to be changed to only ignore annual bill runs.

It also needs to include the selected financial year in the request payload it sends. This is because for SROC supplementary the financial year is always the current one (minus 6) so is determined in `water-abstraction-system`.

Finally, we want to put this behind a feature flag. Fortunately, the existing code already checks for the existence of a feature flag. This was built for SROC supplementary which is now live so that flag is redundant. We just need to rename it and use it to put SROC 2PT behind a flag.


Jozzey:
Put back to draft as I need to make some tweaks

Cruikshanks (approved):

PR #2467, opened 2023-11-07
Remove charging forecast reports from UI

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4168

This PR is removing all the unused code that covers the charging forecast reports. Due to the reports generating incorrect data, they are not used. These reports will not be fixed either as there is an alternative reporting solution.


Cruikshanks (approved):
The very best kind of PR!

Jozzey (approved):

PR #2473, opened 2023-11-19
Redirect to new bill run page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4223

We have been working on implementing a new version of the bill page. The current one tries to display all transaction details for all licences on a single page when the bill is selected.

All that work has been done in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system) as part of our ongoing efforts to migrate away from the legacy code. It is in the final throws of testing and will soon be released.

To get to the new pages you have to pass through the existing bill run summary page. What our work has highlighted is this page also leaves a lot to desire! Again, the larger the bill run the slower it is, noticeably so. Though it at least always loads 😁. It also is inconsistent. It has a different layout depending on the bill run type, and one of those depends on tabs which is an incorrect use of the component.

We'd like to remove this remaining 'bump' in our billing user journey so have rebuilt the page in **water-abstraction-system**. So, now if the feature toggle `USE_NEW_BILL_VIEW` is enabled we direct the user to the new page.

---

We also use the opportunity to simplify what is going on in `getBillingBatchRoute()`. At least we think it is simpler and clearer!

### DIFF
--- a/src/internal/modules/billing/lib/routing.js
+++ b/src/internal/modules/billing/lib/routing.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const { featureToggles } = require('../../../config.js')
+
 /**
  * Gets the correct route for the specified batch depending on its
  * current status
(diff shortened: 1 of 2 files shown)


Cruikshanks:
Do not merge until folks have signed off the new bill run page and the changes are merged into **water-abstraction-system**!!

Cruikshanks:
Because the previous version relied on generating a `Map` and then selecting from it to determine which route to use the existing unit tests gave the impression it was fully covered.

Only a couple of the possible scenarios are and our rewrite has exposed this. But we're rolling out our standard argument that this is the legacy code and the effort is better spent trying to move away from it (which this change falls under).

So SonarCloud is right, but we're choosing to ignore it this time.

PR #2491, opened 2024-01-16
Redirect to new bill run page from batch sent page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4223

We have been working on implementing a new version of the bill page. The current one tries to display all transaction details for all licences on a single page when the bill is selected.

All that work has been done in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system) as part of our ongoing efforts to migrate away from the legacy code. It is in the final throws of testing and will soon be released.

Late in testing, we've spotted that the 'send a bill run' success page displays a 'Go to the bill run' link which is currently taking you to the old view.

So, this change applies a similar change we did in [Redirect to new bill run page](https://github.com/DEFRA/water-abstraction-ui/pull/2473), using the feature toggles to determine what link to use.


Jozzey (approved):
👍🏼

PR #2645, opened 2024-11-29
Make bulk upload work for quarterly

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4704

This was some exploratory work by @chris-barrett-ddts to understand how the bulk upload could work for quarterly returns.

It also encompasses some improvements that allow _all_ due returns to be provided in the CSV download, not just the current ones.

> These and the changes in [water-abstraction-service](https://github.com/DEFRA/water-abstraction-service/pull/2661) are needed for this to work.

### DIFF
--- a/src/external/config.js
+++ b/src/external/config.js
@@ -83,7 +83,7 @@ module.exports = {
   },
 
   returns: {
-    showFutureReturns: testMode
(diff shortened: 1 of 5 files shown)


Cruikshanks:
Implemented by #2679

PR #2656, opened 2025-01-20
Add proxies for image asset requests to WRLS

rvsiyad:
We use proxying to send requests from the legacy system to the `water abstraction system` for our new services. We need to also add the routes for proxying images, similar to the proxying of other assets like scss.

This PR adds proxying for images to the system proxy routes.

### DIFF
--- a/src/shared/modules/system-proxy/routes.js
+++ b/src/shared/modules/system-proxy/routes.js
@@ -77,7 +77,7 @@ const routes = [
     },
     config: {
       auth: false,
-      description: 'Proxies CSS asset requests to the Water Abstraction System'
+      description: 'Proxies image asset requests to the Water Abstraction System'
     }
   },
   {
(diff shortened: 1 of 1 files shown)


src/shared/modules/system-proxy/routes.js
```diff
@@ -66,6 +66,48 @@ const routes = [
       description: 'Proxies CSS asset requests to the Water Abstraction System'
     }
   },
+  {
+    method: 'GET',
+    path: '/assets/images/example-excel-01-min.png',
```
rvsiyad, line 71: We originally tried to use ` path: '/assets/images/{path*}'`, which I believed had worked, but after hard refreshing the page, the images were broken. As a result, I have returned to using the hard coded values for the static images on the [multiple readings guidance page](https://eaflood.atlassian.net/browse/WATER-4812).

Cruikshanks (approved):

PR #2665, opened 2025-02-12
Update notifications setup redirect path

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4899

As part of the work to implement notifications in the system repo we have multiple journeys for notifications.

This change updates the paths for invitations and ad-hoc returns.

The paths point to the same endpoint with a different query string highlighting which notification type has been chosen.


Cruikshanks (changes requested):
It's a niggly thing, but could we change `/system/notifications/setup?notification=` to `/system/notifications/setup?journey=`.

I just feel it then ties it in with the value in the session 🤷

Cruikshanks (approved):

PR #2723, opened 2025-07-10
Update notice setup journey paths

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5143

We need to refactor the existing notice setup journey structure to accommodate the coming adHoc invitations and paper forms.

This change updates the previous 'journey' to 'adHoc' or 'standard'.

The previously known journeys of 'invitations' and 'reminders' are now notice types. This updates fits in with upcoming changes where the user will select which notice type from the select notice type page (currently selected via a link)


jonathangoulding:
System changes - https://github.com/DEFRA/water-abstraction-system/pull/2181

Cruikshanks (approved):

PR #2732, opened 2025-08-14
Change top menu to profile details

ben-sagar:
https://eaflood.atlassian.net/browse/WATER-5171

Rename the current contact information top menu to profile details and the page title.

Move the whole thing from `water-abstraction-ui` into `water-abstraction-system`.

Put this behind a feature flag in `water-abstraction-ui`.

### DIFF
--- a/src/internal/lib/view/proposition-links.js
+++ b/src/internal/lib/view/proposition-links.js
@@ -11,7 +11,6 @@ const createPropositionLink = (label, path, id) => {
   return createLink(label, path, id, { id })
 }
 
-const contactLink = _contactLink()
 const changePasswordLink = createPropositionLink('Change password', '/account/update-password', 'change-password')
 const signoutLink = createPropositionLink('Sign out', '/signout', 'signout')
 
@@ -29,7 +28,7 @@ const getPropositionLinks = (request) => {
   const links = []
 
   if (isHofOrRenewalNotifications(request)) {
-    links.push(contactLink)
+    links.push(_contactLink())
   }
   links.push(changePasswordLink, signoutLink)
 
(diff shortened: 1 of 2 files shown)


src/internal/lib/view/proposition-links.js
```diff
@@ -10,7 +11,9 @@ const createPropositionLink = (label, path, id) => {
   return createLink(label, path, id, { id })
 }
 
-const contactLink = createPropositionLink('Contact information', '/contact-information', 'contact-information')
+const contactLink = useNewProfilePage
+  ? createPropositionLink('Profile details', '/system/users/me/profile-details', 'profile-details')
+  : createPropositionLink('Contact information', '/contact-information', 'contact-information')
```
Cruikshanks, line 16: TBH, it's the legacy code, so I'm not too fussed in this case. But we have a convention (another unwritten one 😩 😬 ) when it comes to the ternary operator in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system).

If you can get your statement on one line, then the ternary is fine. If you have to split it because you've hit the 120 line limit, then it needs something else.

Typically, in this case, I'd suggest creating a 'named' function.

```javascript
const contactLink = _contactLink()

// ...

function _contactLink() {
  if (featureToggles.useNewProfilePage) {
    return createPropositionLink('Profile details', '/system/users/me/profile-details', 'profile-details')
  }

  return createPropositionLink('Contact information', '/contact-information', 'contact-information')
}
```

Cruikshanks (changes requested):
It's only the env var rename I'd push for a change (but by all means challenge that decision!)

The comment about when to use the ternary operator was more for reference.

PR #2833, opened 2026-06-18
Add .npmrc with supply chain security settings

jonathangoulding:
Adds a root `.npmrc` with `save-exact=true`, `ignore-scripts=true`, and `min-release-age=7`.
Pins installed versions exactly, disables install scripts, and sets a minimum release age for new dependency versions to reduce exposure to compromised npm releases.


jonathangoulding:
Closing, no longer needed.

